PR #7, opened 2016-09-21
Support integer, atom or strings as aggregate_uuid in Commanded.Aggregates.Registry.open_aggregate(aggregate_module, aggregate_uuid)

andrzejsliwa:
lets integer, atom or strings as aggregate_uuid in 
```
Commanded.Aggregates.Registry.open_aggregate(aggregate_module, aggregate_uuid)
```

using other values than string raising low level error on postgrex driver about text type


slashdotdash:
Great, thank you

PR #8, opened 2016-09-24
Fix misspell in supervisor example

astery:


slashdotdash:
Thanks for catching and fixing this.

astery:
Glad to help all of us.

PR #10, opened 2016-09-26
Don't persist an aggregate's pending events when executing a command returns an error

### LINKED ISSUE #9: Don't persist an aggregate's pending events when executing a command returns an error
When a command is dispatched it is routed to the single registered command handler by the configured router. The aggregate is loaded, or created as necessary. The handler executes a function on the aggregate passing in the received aggregate state. 

The aggregate and command handler both return the aggregate state on success. The pending events are persisted to the event store.

Should the command fail, an `{:error, ... }` tuple should be returned. This would prevent the aggregate state and pending events from being persisted.

slashdotdash:
Don't persist an aggregate's pending events when executing a command returns an error.

An Aggregate should enforce domain rules by returning an error tuple instead of its state on failure. 

- An aggregate function must return `{:ok, aggregate}` on success.
- An aggregate function should return `{:error, reason}` on failure.

The example bank account aggregate returns `{:error, :account_already_open}` should an attempt be made to open an existing account.

```elixir
defmodule Commanded.ExampleDomain.BankAccount do
  def open_account(%BankAccount{state: %{is_active?: true}} = account, %OpenAccount{}) do
    {:error, :account_already_open}
  end

  def open_account(%BankAccount{state: %{is_active?: false}} = account, %OpenAccount{account_number: account_number, initial_balance: initial_balance}) when is_number(initial_balance) and initial_balance > 0 do
    account = 
      account
      |> update(%BankAccountOpened{account_number: account_number, initial_balance: initial_balance})

    {:ok, account}
  end
end
```

The error will be returned to the caller when the command is dispatched.

```elixir
# open account should initially succeed
:ok = Router.dispatch(%OpenAccount{account_number: "acc1"})

# attempting to open same account should error
{:error, :account_already_open} = Router.dispatch(%OpenAccount{account_number: "acc1"})
```

Fixes #9.

### DIFF
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -12,8 +12,6 @@ defmodule Commanded.Aggregates.Aggregate do
 
   defstruct aggregate_module: nil, aggregate_uuid: nil, aggregate_state: nil
 
-  @command_retries 3
-
   def start_link(aggregate_module, aggregate_uuid) do
     GenServer.start_link(__MODULE__, %Aggregate{
       aggregate_module: aggregate_module,
@@ -55,7 +53,7 @@ defmodule Commanded.Aggregates.Aggregate do
   Execute the given command, using the provided handler, against the current aggregate state
   """
   def handle_call({:execute_command, command, handler}, _from, state) do
-    {reply, state} = execute_command(command, handler, state, @command_retries)
+    {reply, state} = execute_command(command, handler, state)
 
     {:reply, reply, state}
   end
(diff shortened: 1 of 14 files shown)


lib/commanded/aggregates/aggregate.ex
```diff
@@ -77,32 +77,57 @@ defmodule Commanded.Aggregates.Aggregate do
...
     expected_version = version
 
-    aggregate_state = handler.handle(aggregate_state, command)
+    with {:ok, aggregate_state} <- handle_command(handler, aggregate_state, command),
+         {:ok, aggregate_state} <- persist_events(aggregate_state, expected_version)
+      do {:ok, %Aggregate{state | aggregate_state: aggregate_state}}
+    else
+      {:error, :wrong_expected_version} -> retry_command(command, handler, state, retries - 1)
+      {:error, _reason} = reply -> {reply, state}
+    end
```
andrzejsliwa: looks good!

lib/commanded/aggregates/aggregate.ex
```diff
@@ -77,32 +77,57 @@ defmodule Commanded.Aggregates.Aggregate do
...
+  defp persist_events(%{uuid: aggregate_uuid, pending_events: pending_events} = aggregate_state, expected_version) do
     correlation_id = UUID.uuid4
     event_data = Mapper.map_to_event_data(pending_events, correlation_id)
 
-    EventStore.append_to_stream(aggregate_uuid, expected_version, event_data)
+    case EventStore.append_to_stream(aggregate_uuid, expected_version, event_data) do
+      :ok ->
+        # clear pending events after successful store
+        {:ok, %{aggregate_state | pending_events: []}}
+      {:error, :wrong_expected_version} = reply ->
```
andrzejsliwa: I wonder if we could have here strong matching:

{:ok, state} = EventStore.append_to_stream(...)

to fail immediately, otherwise *with* will suppress error and only logger will be the evidence of problem. I personally prefer  matching result for success case only and fail in place where errors occurs. In this case this will stop the process with information about state and reason. This is all about using OTP to have less defensive code and handle only happy cases. 

Defensive code with *with* make sense for me in places where you need to handle user input/output or in business logic, but even then I'm always thinking if handling of not success case can't be handled on level of supervision tree instead of defensive code.

lib/commanded/aggregates/aggregate.ex
```diff
@@ -77,32 +77,57 @@ defmodule Commanded.Aggregates.Aggregate do
...
+  end
 
-        # retry command
-        execute_command(command, handler, state, retries - 1)
+  defp handle_command(handler, aggregate_state, command) do
+    case handler.handle(aggregate_state, command) do
+      {:error, reason} = reply ->
+        Logger.debug(fn -> "failed to handle command due to: #{inspect reason}" end)
+        reply
+      aggregate_state -> {:ok, aggregate_state}
```
andrzejsliwa: what do you think about returning {:ok, state} | {:error, reason} also from handler?

slashdotdash: That's a good suggestion. Being explicit about handlers returning `:ok` and the aggregate state on success makes sense. I'll look at making this change.

lib/commanded/aggregates/aggregate.ex
slashdotdash: Are you suggesting to remove the `{:error, :wrong_expected_version}` match and only match on :ok`.  This would crash the aggregate process if persisting the pending events fails due to version mismatch. That would follow the 'let it crash' idiom.

I haven't yet thought too much about retrying command handling on failure. I'd like this to be client configurable (e.g. 5 retries, with exponential back-off).

andrzejsliwa: > Are you suggesting to remove the {:error, :wrong_expected_version} match and only match on :ok. This would crash the aggregate process if persisting the pending events fails due to version mismatch. That would follow the 'let it crash' idiom.

yes, exactly. GenRetry is good example, can wrap in function this call
*{:ok, state} = EventStore.append_to_stream(...) ...*

PR #13, opened 2016-09-30
Move event assertions from tests to library

astery:
Then they can be used in user application tests


slashdotdash:
Useful change, thanks.

PR #17, opened 2016-10-10
forward error of EventStore.append_to_stream to CommandDispatcher

andrzejsliwa:
forward error of EventStore.append_to_stream to CommandDispatcher


andrzejsliwa:
its WIP, I will add missing tests for it

slashdotdash:
@andrzejsliwa There's an existing test that asserts the current behaviour is to crash the aggregate process when an error is encountered (e.g. wrong expected version). This will force the aggregate to fetch its events from storage the next time a command is received for that aggregate instance. If the aggregate process doesn't crash, then it will never recover from this error.

I think command dispatching should be retried on error. It should be configured in the router that uses the `dispatch` macro.

PR #22, opened 2016-10-31
Fix pending aggregates restarts: Supervisor restarts aggregate, that loses command routing and remains pending

henry-hz:
When we start a new aggregate, and kill it, supervisor immediately restart it with the last state, replaying all the events. But when sending a new Command to the same aggregate UUID, a new aggregate is created, and updated, so we end up with two aggregates on memory, one unusable (and consumed resources to be restarted) and a new fresh one, that was started by the last command request. Using :observer.start can help to track their state and see what is going on. Pls, let me know if you need more details to reproduce it.

### DIFF
--- a/lib/commanded/aggregates/supervisor.ex
+++ b/lib/commanded/aggregates/supervisor.ex
@@ -18,7 +18,7 @@ defmodule Commanded.Aggregates.Supervisor do
 
   def init(_) do
     children = [
-      worker(Commanded.Aggregates.Aggregate, [], restart: :transient),
+      worker(Commanded.Aggregates.Aggregate, [], restart: :temporary),
     ]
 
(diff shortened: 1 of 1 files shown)


henry-hz:
Restart Strategies:
    :permanent - the child process is always restarted
    :temporary - the child process is never restarted (not even when the supervisor’s strategy is :rest_for_one or :one_for_all)
    :transient - the child process is restarted only if it terminates abnormally, i.e., with an exit reason other than :normal, :shutdown or {:shutdown, term}

If we leave  as transient, the process will be pending in case of any abnormal termination.

slashdotdash:
Thanks for the pull request & fix to this issue. 

The same issue applies to the process manager supervisor as it is also using the `:transient` restart strategy.

PR #47, opened 2017-01-06
Eventstore behaviour / Integration of geteventstore.com

sharksdontfly:
This fork integrates geteventstore.com (via extreme). Event store usage in commanded is replaced by a behaviour. It has one limitation though, when using `extreme` event store ack'ing of events is not implemented. To run all tests for `eventstore` or `extreme` event store you have to execute `MIX_ENV=test mix clean && COMMANDED_EVENT_STORE_ADAPTER=Commanded.EventStore.Adapters.EventStoreEventStore mix test` or `MIX_ENV=test mix clean && COMMANDED_EVENT_STORE_ADAPTER=Commanded.EventStore.Adapters.ExtremeEventStore mix test` respectively. Testing `extreme` event store takes considerably longer because each time the event store is resetted a docker container is restarted. When using the `extreme` event store one has to configure extreme and additionally provide a value for `:streams_prefix` (see `config/test.exs`). So if you configure `bankaccount` here and an event for account `ACC123` is written to the event store it will be written to the stream `bankaccount-ACC123`.

Additionally there is a module which one can `use` to customize event names, it works only when using the integrated `Commanded.Serialization.JsonSerializer`. E.g.:

```
defmodule Events do
  alias Commanded.Serialization.JsonSerializer.EventName

  defmodule MoneyTransferRequested do
    use EventName, "money-transfer-requested"

    defstruct [transfer_uuid: nil, debit_account: nil, credit_account: nil, amount: nil]
  end
end
```


slashdotdash:
Thanks @sharksdontfly. Great effort, I look forward to getting it merged in.

sharksdontfly:
@slashdotdash Thank you. Tell me if something needs to be changed.

henry-hz:
@sharksdontfly what do you mean by "It has one limitation though, when using extreme event store ack'ing of events is not implemented." ?  is this limitation an extreme limitation ? if not, what should be the way to implement it ?

sharksdontfly:
@henry-hz slashdotdash/EventStore requires events to be acked and this is how EventStore provides a back pressure mechanism. As far as I know no such thing exists in Extreme, so when the Extreme adapter is used no back pressure mechanism is available. Concerning implementation see the last text block here: https://github.com/slashdotdash/eventstore#ack-received-events.

henry-hz:
@sharksdontfly seems that in [this](https://github.com/exponentially/extreme/blob/a6515020a7b79ae41680f6ae9dc86c9fa3d72631/test/listener_test.exs) test file, the test "Listener reads existing unacked events", implements back-pressure keeping the acked events in the small DB below. It should be straightforward to implement this in commanded. 

```
defmodule DB do
    def start_link, 
      do: Agent.start_link(fn -> %{} end, name: :db) 

    def get_last_event(listener, stream), 
      do: Agent.get(:db, fn state -> Map.get(state, {listener, stream}, -1) end)

    def ack_event(listener, stream, event_number), 
      do: Agent.update(:db, fn state -> Map.put(state, {listener, stream}, event_number) end)

    def in_transaction(fun), do: fun.()
end
```

sharksdontfly:
@henry-hz I dont't think `DB` module provides for a back pressure mechanism here. It is a mock for a DB, which keeps track of which events have been processed so far and therefore allows to resume processing of events in case of Listener process crash. It does not prevent Listener mailbox to be flooded with incoming events. @slashdotdash Is a back pressure mechanism required for this pull request to be accepted?

slashdotdash:
@sharksdontfly I'd prefer to have both adapters functionality compatible. 

Do you think it will be possible to add the ack behaviour to the Get Event Store adapter?

Back pressure is most needed when you add a new event handler that needs to replay all events from the beginning of time. This scenario could quickly overwhelm the event handler process if it cannot keep up.

sharksdontfly:
@slashdotdash @henry-hz I have now implemented back pressure for the 'get event store' adapter.

slashdotdash:
@sharksdontfly Thanks the for update and making those additional changes.

henry-hz:
@slashdotdash thanks for the follow up and adding back pressure

henry-hz:
hi @slashdotdash , pls let me know if you are going to merge this PR in the next weeks
thanks

slashdotdash:
@henry-hz Have you had a chance to try these changes out? Let me know how it goes.

slashdotdash:
I'm splitting this pull request up into separate issues: #49, #50, #51.

This is so that each storage adapter - to support either the PostgreSQL event store or Greg's geteventstore - will reside in its own project & Git repo. Allowing consumers of Commanded to reference and install only the single storage adapter they wish to use.

henry-hz:
Hi @slashdotdash , I hope to check soon... I will keep you updated. 
Splitting components seems a good idea.

slashdotdash:
Closing as this pull request has been included in and replaced by #55.

PR #53, opened 2017-04-07
Fix having nil in metadata

olafura:
When you have nil in the metadata it fails because it's trying to
merge a nil with a map.


olafura:
An other solution is to not allow nil to be in metadata.

slashdotdash:
@olafura Thanks for the pull request.

PR #54, opened 2017-04-16
WIP: Wires in support for Causation IDs

harukizaemon:
TODO: Not sure how to do this given I don't have a good feeling for the framework principles.

Correlation ID: Typically, a ProcessManager would copy the Correlation ID from the event.  This allows the Correlation ID to propagate and makes it easy to isolate a chain of events.  In the case of a user-driven command such as via a UI the Correlation ID would be synthesized as now.

Causation ID: Typically, a ProcessManager would use the Event ID as the Causation ID.  This allows the Causation ID to propagate and makes it esy to isolate specific cause-effect relationships.  In the case of a user-driven comand uch as via a UI the Causation ID would either be left null, or in some frameworks assigned the Command ID if there is one.  However, as it stands, Event IDs are integer, rather than UUIDs making this difficult.

### DIFF
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -142,7 +142,29 @@ defmodule Commanded.Aggregates.Aggregate do
 
         updated_state = apply_events(aggregate_module, aggregate_state, pending_events)
 
-        :ok = persist_events(pending_events, aggregate_uuid, expected_version)
+        # TODO: Not sure how to do this given I don't have a good feeling for
+        #       the framework principles.
+        #
+        # Correlation ID:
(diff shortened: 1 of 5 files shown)


slashdotdash:
I'm not sure how best to solve the problem of events using an integer `event_id` rather than a UUID. 

A couple of options I can immediately think of:

1. Change `event_id` to a UUID - persisted as a string - as per correlation_id and causation_id. Add an `event_number` field to the events table to act as the global event sequence number.
2. Use a naming strategy for correlation_id / causation_id such as `event-1234`. 

Option 1. requires most rework and adds additional storage requirements to the events table. Every event would include an additional UUID property. Second option requires minimal code change. Any thoughts?

slashdotdash:
Closing this issue as it requires further work.

Issue #43, opened 2016-12-06
Shutdown idle aggregate processes


slashdotdash:
An aggregate process that has been idle for a configured period of time should be shutdown.

`GenServer`s have built in support for timeouts. You include a timeout value when replying to a `handle_call` or `handle_cast` request. If no subsequent requests are made to your GenServer in the timeout period it sends `:timeout` to `handle_info`. You can then shutdown the process.

henry-hz:
maybe add a mechanism to generate an event before shutdown, so sagas will be able to reverse any long transaction ?

Papipo:
It doesn't make sense that aggregates emit events because they haven't been receiving commands lately.

henry-hz:
yep, agree.  there are two reasons to shutdown, or the business process was finished, or we need more memory, and that aggregate has been sparsely used. for the first, the PM could shutdown them sending stop commands, but for the former, the solution above is fine. an algorithm could calculate the desired "cache" memory, and sets the best timeout based on memory needs

Papipo:
I guess that this is one of those cases that depends on the use case. Providing configurable timeouts would probable suffice. It's just a matter of finding a good value for your application.

bbhoss:
Hibernating after a timeout could also be a good idea, since it will reduce memory to be as small as possible. You could probably have a timeout handler that registers (probably in ETS/Registry) the process somewhere as "in hibernation", with the current time, then have some sort of reaper that decides when to kill/shutdown processes. This is a common issue in actor languages, and why Pony was built to allow garbage collecting the actors themselves, which seems like a pretty rare/one of a kind ability.

slashdotdash:
@bbhoss GenServer already has support for handling timeouts. You can return a `timeout` (in milliseconds) as part of the reply to `handle_cast` and `handle_call` requests. If there are no subsequent requests within the given timeout period, the GenServer will receive a `:timeout` message and can be stopped. 

It *just* needs to be implemented for the aggregate GenServer.

bbhoss:
I understand that, what I'm proposing is in the timeout message handler, we register as hibernated, then return a `:hibernate` tuple. Then a configurable reaper could evict the processes later.

slashdotdash:
Done in #57.

slashdotdash:
Published to Hex as [v0.11.0](https://hex.pm/packages/commanded/0.11.0)

PR #57, opened 2017-06-02
Shutdown idle aggregate processes.

astery:
I propose implementing the following behavior to manage aggregate lifetime. The main intention was that aggregate life controller should be separated entity from domain logic.

```elixir
defmodule ShopLifespan do
  @behaviour Commanded.Aggregates.AggregateLifespan

  # It specifies how long aggregate should live after processing command
  # @callback after_command(command) :: integer() | :infinity | :hibernate
  def after_command(%StartShopWork{}), do: 3600 * 1000 * 8
  def after_command(%StopShopWork{}), do: :hibernate
  def after_command(%CloseShop{}), do: 0
end
```

It should be specified in router:

```elixir
defmodule CommandRouter do
  use Commanded.Commands.Router

  dispatch [
     StartShopWork,
     StopShopWork,
     CloseShop,
  ], to: MyAggregate, lifespan: MyAggregateLifespan, identity: :aggregate_id
end
```

### DIFF
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -23,6 +23,12 @@ defmodule Commanded.Aggregates.Aggregate do
     aggregate_version: 0,
   ]
 
+  defmodule DefaultLifespan do
+    @behaviour Commanded.Aggregates.AggregateLifespan
+
+    def after_command(_), do: :infinity
+  end
+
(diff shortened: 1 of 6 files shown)


slashdotdash:
@astery This looks like a great change, thanks for taking the time to implement #43.

astery:
I'm glad to hear that, thanks.

PR #65, opened 2017-07-22
Allow compiling in other registry providers

brentonannan:
Gday @slashdotdash,

I'd like to use commanded across multiple nodes, the first step of which would be to be able to run it with a global process registry. This PR allows you to compile commanded using a registry provider other than `Registry` by specifying the module in config as `:registry_provider`. It will should work with any registry that supports `:via` tuples (thus the changes to the tests - `lookup` isn't one of the callbacks required to support them), while still compiling with `Registry` by default.

I was toying with the idea of writing a `Commanded.Registry.Provider` behaviour, with a default implementation using `Registry` as `Commanded.Registry` but this was the simpler option and I wanted to get feedback on the idea first.

I have a branch [here](https://github.com/brentonannan/commanded/tree/test/compile_with_syn_as_registry_provider) which is green for all tests when using [syn](https://github.com/ostinelli/syn).

### DIFF
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -13,6 +13,7 @@ defmodule Commanded.Aggregates.Aggregate do
   alias Commanded.Aggregates.Aggregate
   alias Commanded.Event.Mapper
 
+  @registry_provider Application.get_env(:commanded, :registry_provider, Registry)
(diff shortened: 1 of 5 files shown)


slashdotdash:
@brentonannan Neat, thanks for the pull request. I'm in favour of taking simple steps towards distributed running.

PR #75, opened 2017-08-31
Adjust documentation to make Process Managers work

leifg:
If you start it with only the keyword list, the application start stops with `        ** (FunctionClauseError) no function clause matching in Keyword.merge/2` because the keyword list is not the first argument but the whole list of arguments.


slashdotdash:
Thanks for the fix @leifg.

PR #78, opened 2017-09-15
Fixed dead links in usage guide

JoanZapata:


slashdotdash:
@JoanZapata The guides are used for Hex docs, so the links reference pages hosted there. That's why they don't work on GitHub.

Here's the [Usage](https://hexdocs.pm/commanded/usage.html) page on Hex docs for example.

It's probably easiest just to remove the links so they aren't broken on GitHub. Making them full URLs, not relative links, would break when viewing documentation for older versions on Hex docs. It's a bit of a catch-22 situation!

JoanZapata:
Oh I see, I missed that part sorry. Maybe a third option would be to reference the Hex docs from the `README`, so that people don't end up clicking from here like I did.

Thanks for your review!

PR #88, opened 2017-10-04
Remove unnecessary warning when a process managers stops itself

JoanZapata:
A normal stop is already logged in `handle_event/2` with a debug level.


slashdotdash:
Thanks for the pull request @JoanZapata.

PR #90, opened 2017-10-05
Fail `assert_receive_event` if `event_type` isn't a known module

febeling:
`assert_receive_event` waits for an event of a specific type. But it
doesn't recognize if the given event type isn't valid, e.g. because it
hasn't been alias or has a typo. That way a test that should succeed
if the module had been aliased fails.

This change ensures the type is a known one or else raises an
assertion exception which indicates the fact.


slashdotdash:
Nice change @febeling, I've been caught out by forgetting to alias an event numerous times.

febeling:
Thanks, I'm happy you find it useful!

PR #91, opened 2017-10-08
Fix typo in Changelog

davich:


slashdotdash:
Good catch, thanks @davich.

PR #103, opened 2017-10-30
Let middlewares modify metadata before dispatch

JoanZapata:
This is useful for middleware like [commanded-audit-middleware](https://github.com/commanded/commanded-audit-middleware/), which would be able to put the command uuid inside metadata automatically.


slashdotdash:
Thanks @JoanZapata.

Issue #56, opened 2017-05-15
Explicit aggregate `create` functions


slashdotdash:
A function that creates an aggregate root is currently the same as any function expecting an existing aggregate instance. 

You must use some field from the aggregate's state to determine whether it is a new (default state) or existing (populated from the first raised event) instance.

### Current: Open account with default `%BankAccount{}` state

The `execute/2` function receives default state when no events exist for the aggregate:

```elixir
defmodule Commanded.ExampleDomain.BankAccount do
  defstruct [
    account_number: nil,
    balance: 0,
    state: nil,
  ]

  def execute(%BankAccount{state: nil}, %OpenAccount{account_number: account_number, initial_balance: initial_balance})
    when is_number(initial_balance) and initial_balance > 0
  do
    %BankAccountOpened{account_number: account_number, initial_balance: initial_balance}
  end
end
```

### Proposal: Open account with no state

An `execute/1` function would be called when there are no events for the given aggregate (i.e. we are creating a new instance):

```elixir
defmodule Commanded.ExampleDomain.BankAccount do
  defstruct [
    account_number: nil,
    balance: 0,
    state: nil,
  ]

  def execute(%OpenAccount{account_number: account_number, initial_balance: initial_balance})
    when is_number(initial_balance) and initial_balance > 0
  do
    %BankAccountOpened{account_number: account_number, initial_balance: initial_balance}
  end
end
```

This makes it explicit that certain aggregate function(s) expect to create a new instance. Pattern matching will cause an exception when attempting to create more than one instance (e.g. open an account for the same account number).

drozzy:
How will this translate to non-execute version? Or will it just stay the same (which is ok I think)?

slashdotdash:
@drozzy Good catch, I think that also needs to change, proposal as below.

### Current: Command handler receives default `%BankAccount{}` state

```elixir
defmodule OpenAccountHandler do
  @behaviour Commanded.Commands.Handler

  def handle(%BankAccount{} = aggregate, %OpenAccount{account_number: account_number, initial_balance: initial_balance}) do
    BankAccount.open_account(aggregate, account_number, initial_balance)
  end
end
```

### Proposal: Command handler receives no state

When there are no persisted events for the aggregate.

```elixir
defmodule OpenAccountHandler do
  @behaviour Commanded.Commands.Handler

  def handle(%OpenAccount{account_number: account_number, initial_balance: initial_balance}) do
    BankAccount.open_account(account_number, initial_balance)
[...]

PR #106, opened 2017-11-03
Improve usage example (re-order pattern matches)

fdbeirao:
I (believe!) understand how pattern matching works in Elixir (top to bottom, eager). I was looking into this example and I couldn't grasp how the previous code would "ensure account has not already been opened". 

I re-ordered some of the code. It basically has all the invariants before the actual happy path, plus the `not is_nil()` guard.

In my head, this code that I am proposing would make more sense. Please, do correct me if I am wrong.
 
⚠️ **I have not tested this code.** This was written more as "pseudo-code", from some examples on Elixir Guards.

### DIFF
--- a/guides/Usage.md
+++ b/guides/Usage.md
@@ -39,10 +39,11 @@ Here's an example bank account opening feature built using Commanded to demonstr
 
       # public command API
 
-      def execute(%BankAccount{account_number: nil} = account, %OpenBankAccount{account_number: account_number, initial_balance: initial_balance})
-        when initial_balance > 0
+      # ensure account has not already been opened
+      def execute(%BankAccount{account_number: existing_account_number} = account, %OpenBankAccount{}) 
+        when not is_nil(existing_account_number)
       do
-        %BankAccountOpened{account_number: account_number, initial_balance: initial_balance}
+        {:error, :account_already_opened}
       end
 
       # ensure initial balance is never negative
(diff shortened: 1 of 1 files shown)


slashdotdash:
Thanks for the feedback, I'd make two suggestions here: 

- `account_number` in the OpenBankAccount  command can never be `nil` as this is used to route the command to the aggregate instance, so the check is unnecessary. Attempting to dispatch a command with a `nil` aggregate identity will return `{:error, :invalid_aggregate_identity}`.
- My preference is to order functions by the success case first and handle errors afterwards. It's also possible to exclude the error matches entirely and allow a runtime error when an invalid command or data is given to the aggregate (following the Erlang "let it crash" philosophy). 

There's a proposal to alter how aggregate's respond to commands when they are being created (#56). I think that approach might help to make it more explicit that the command must be executed against a new aggregate, without requiring the state field check for `nil`.

fdbeirao:
> * Attempting to dispatch a command with a nil aggregate identity will return `{:error, :invalid_aggregate_identity}`.

I was not aware of this fact. It addresses the problem correctly. 👍 

> * My preference is to order functions by the success case first and handle errors afterwards.

I have **not** used Commanded *yet*. All I have is a gut feeling that this would create some sort of "redundancy" in my happy path guards. Redundancy, because I do want the useful error reason checks. I like the let it crash philosophy, but for aggregates I really appreciate having some more "business/domain specific error", rather than a technical "No pattern matched" or something. 

My instinct would be to put all error/edge conditions first, and the happy path last. Thinking about it, I believe this hunch comes from the way you implement recursive functions with pattern matching. Of course I might be wrong hehe.

----

Back to the point that brought me to this PR in the first place, I can't wrap my head around this particular piece of code:
```elixir
  # ensure account has not already been opened
  def execute(%BankAccount{account_number: nil} = account, %OpenBankAccount{}) do
    {:error, :account_already_opened}
  end
```

Given the comment "ensure account has not already been opened", for me, only an implementation such as this would make sense:

```elixir
  # ensure account has not already been opened
  def execute(%BankAccount{account_number: existing_account_number} = account, %OpenBankAccount{})
    when not is_nil(existing_account_number)
  do
    {:error, :account_already_opened}
  end
```

Either that, or I am missing something quite obvious. 

Thank you for your time @slashdotdash 👍

slashdotdash:
>  I can't wrap my head around this particular piece of code:

```elixir
# ensure account has not already been opened
def execute(%BankAccount{account_number: nil} = account, %OpenBankAccount{}) do
  {:error, :account_already_opened}
end
```

Ah, I understand your confusion now. The above code is wrong, it should be:

```elixir
# ensure account has not already been opened
def execute(%BankAccount{}, %OpenBankAccount{}) do
  {:error, :account_already_opened}
end
```

That's my fault, a copy & paste error, oops.

PR #109, opened 2017-11-08
Support distributed dispatch consistency 

slashdotdash:
Allow command dispatch with `consistency: :strong` to be used when Commanded is distributed amongst a cluster of nodes

### DIFF
--- a/lib/commanded/subscriptions/subscriptions.ex
+++ b/lib/commanded/subscriptions/subscriptions.ex
@@ -2,6 +2,7 @@ defmodule Commanded.Subscriptions do
   @moduledoc false
 
   use GenServer
+  use Commanded.Registration
 
   alias Commanded.EventStore.RecordedEvent
   alias Commanded.Subscriptions
@@ -14,7 +15,7 @@ defmodule Commanded.Subscriptions do
   ]
 
   def start_link(_arg) do
-    GenServer.start_link(__MODULE__, %Subscriptions{}, name: __MODULE__)
+    Registration.start_link(__MODULE__, __MODULE__, %Subscriptions{})
   end
 
   @doc """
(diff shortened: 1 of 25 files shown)


lib/commanded/subscriptions/subscriptions.ex
```diff
@@ -119,30 +130,30 @@ defmodule Commanded.Subscriptions do
...
-    {:reply, :ok, state}
+    {:noreply, state}
   end
 
-  def handle_cast({:ack_event, name, stream_uuid, stream_version, pid}, %Subscriptions{streams_table: streams_table, subscriptions_table: subscriptions_table, started_at: started_at} = state) do
-    # track insert date as ms since the subscripions process was started to support expiry of stale acks
+  def handle_info({:ack_event, name, stream_uuid, stream_version, pid}, %Subscriptions{streams_table: streams_table, subscriptions_table: subscriptions_table, started_at: started_at} = state) do
+    # track insert date as milliseconds since the subscriptions process was
+    # started to support expiry of stale acks
     inserted_at_epoch = NaiveDateTime.diff(now(), started_at, :second)
```
fdbeirao: Is it normal for the comment to say "track (...) as milliseconds", and then the `diff` function is called with unit ``:second`?

slashdotdash: No, the comment is a lie! I'll fix it, thanks.

test/process_managers/support/error/error_handling_process_manager.ex
```diff
@@ -0,0 +1,95 @@
...
+  defp record_attempt(context) do
+    Map.update(context, :attempts, 1, fn attempts -> attempts + 1 end)
+  end
+
+  defp reply_failure(context) do
+    reply({:error, :failed, context})
+  end
+
+  defp reply(message) do
+    reply_to = Agent.get({:global, ErrorHandlingProcessManager}, fn reply_to -> reply_to end)
```
fdbeirao: I believe here you could use a similar `@constant` for `{:global, ErrorHandlingProcessManager}`, like you did [here](https://github.com/commanded/commanded/pull/109/files#diff-5d66fbd4d07acba5c44eebee23801a57R5).

PR #115, opened 2017-11-29
Update README.md

dustinfarris:
Limitations section does not seem to exist


slashdotdash:
Thanks for catching that, I removed the limitations section as it was no longer relevant but forgot to remove it from the contents.

PR #127, opened 2018-01-20
Add new values for expected_version

leifg:
The eventstore backend allows to skip the expected version check: (https://github.com/commanded/eventstore/blob/2d031e73a668fb518fc96f0670ccc6ed1deb5bc8/lib/event_store.ex#L44)

This PR will extend this feature to the InMemory adapter.

PR for [eventstore-adapter](https://github.com/commanded/commanded-eventstore-adapter) will follow


slashdotdash:
This requires making a breaking change to the [EventStore](https://github.com/commanded/eventstore) library since it only supports returning an `:ok` response when appending events.

I'm somewhat hesitant to make this change without a strong need to.

PR #129, opened 2018-01-24
Fix typo in Read Model Projections.md

Trevoke:


slashdotdash:
Thanks again @Trevoke.

PR #145, opened 2018-02-16
add assert_correlated test helper

Trevoke:
Like other helpers, can be brought in with `import Commanded.Assertions.EventAssertions`.

This assertion ensures that:
1. Both events, each matching their respective predicates, are received
2. Both events share the same correlation id

Usage sample follows.

Note that in this example I am using the same identifier to identify the two events out of convenience. 

```elixir
identifier = UUID.uuid4()
assert_correlated(
  EventTypeA, &(&1.identifier_a == identifier),
  EventTypeB, &(&1.identifier_b == identifier)
)
```

### DIFF
--- a/lib/commanded/assertions/event_assertions.ex
+++ b/lib/commanded/assertions/event_assertions.ex
@@ -41,6 +41,19 @@ defmodule Commanded.Assertions.EventAssertions do
     end)
   end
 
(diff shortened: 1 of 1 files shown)


slashdotdash:
Thanks @Trevoke.

PR #146, opened 2018-02-16
add guide on testing

Trevoke:
I never remember to go to the wiki, and I think testing should be a first-class citizen.

This is my first pass on making testing code that uses commanded more approachable. All feedback welcome. :)

(it looks like this: https://github.com/Trevoke/commanded/blob/add-testing-guide/guides/testing.md )


slashdotdash:
Thank you @Trevoke.

PR #147, opened 2018-02-23
EventStore adapter tests: Remove flakiness

jfornoff:
The ES adapter tests could fail when the snapshots coincidentially were
created in different minutes, since the seconds were just reset to 0.

This commit updates the logic to consider the difference between
timestamps and asserts a difference no larger than 60 seconds instead.


slashdotdash:
Thanks @jfornoff.

PR #150, opened 2018-03-02
Allows for a default consistency to be set via the env

zamith:
Why:

* In some situations, particularly when testing, we'll want to have
  strong consistency, but that might not be true for the same operation
  in a different environment. So, it would be very helpful to be able to
  switch between types of consistency per environment.

This change addresses the need by:

* Changing the default from being straight up `:eventual` but instead to
  search on the env first and only if nothing is found, then it will
  choose `:eventual`

PS: You'll notice that there are no test for the router change. That is because it is set in a module attribute, which being a compile time evaluated property means I cannot affect from the spec. The only way I could think of testing it, would be to create a new environment just for that purpose, which might be an overkill. What do you think?


slashdotdash:
@zamith Thanks for the pull request.

PR #155, opened 2018-03-17
support custom type provider for EventAssertions .

jccf091:
changed to use Commanded.EventStore.TypeProvider instead of Atom to get event type


slashdotdash:
Thank you @jccf091.

PR #158, opened 2018-03-26
Add Assertion to ListAllDone event

dippysan:
assert_receive_event is missing an assert, so the check for list_uuid never fails.

When learning Commanded, I used this example as a template, then later realised none of my tests were failing and I needed to assert inside function.


slashdotdash:
Good catch, thanks for the pull request @dippysan.

PR #159, opened 2018-03-26
Fix testing link in README

dippysan:
The testing link in README.md linked to guides/Testing.md, which does not exist.

Small change so the link matches the file.


slashdotdash:
I've renamed `testing.md` to `Testing.md` for consistency with the other guides. Thanks for highlighting the issue.

PR #160, opened 2018-03-30
Clear potential adapter/store confusion

pablobm:
I'm new to CQRS and Commanded. Today I was experimenting with Commanded and came across something that threw me off for a bit. The way I read the docs, the difference between an store and its adapter was not clear, at least for the Postgres adapter, which is the one I used.

I'll elaborate. When reading the section "PostgreSQL-based Elixir EventStore", I clicked on the link "EventStore", opening it on a tab and continued reading. When I saw the phrase advising to follow the "Getting started" guide, I assumed it was the one from the store, not the adapter, not realising that the links led to different repositories. Because of this, I installed the store instead of the adapter, and things didn't work until I finally realised my mistake.

I feel that my proposed changes may help highlighting the difference, and make the steps to follow clearer to future readers.

### DIFF
--- a/guides/Choosing an Event Store.md
+++ b/guides/Choosing an Event Store.md
@@ -2,9 +2,9 @@
 
 You must decide which event store to use with Commanded. You have a choice between two existing event store adapters:
 
-- PostgreSQL-based Elixir [EventStore](https://github.com/commanded/eventstore).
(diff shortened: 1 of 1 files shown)


slashdotdash:
Thanks for the feedback and pull request to clarify the adapter/event store confusion @pablobm .

PR #167, opened 2018-04-08
fix link to Greg Young's site.

NigelThorne:
I also took the cheeky step of adding myself as a contributor :)


slashdotdash:
Thanks for the fix Nigel.

PR #168, opened 2018-04-12
Ability to globally override include_execution_result and include_aggregate_version

naps62:
We want to enforce consistent return values in our entire app.
`:include_execution_result` looks exactly like what we want, and
currently fixes two problems for us, so we want to enforce it globally.

I did a similar change to what my colleague zamith already did
previously for another option. however, unlike his case, I didn't find
a good way to test this. Would love to get any input on that front


slashdotdash:
Thanks for the pull request @naps62.

PR #175, opened 2018-06-02
Proposal: InMemory.reset!() for testing purposes

naps62:
# Some background

I'm struggling to get the InMemory store to work with my test suite
Following this [wiki page](https://github.com/commanded/commanded/wiki/In-memory-event-store), my conclusion is that the samples shown there are a rather complex way of rebooting the in-memory event store, which requires rebooting the app itself
(I presume this is because the event store is a dependency of the app, or something like that)

The thing is, needing `test --no-start`, and killing my app between tests is a complex thing that leaks that behaviour to the rest of the test suite. I now have to manually start my app in other test files as well, due to the.

# My proposal

Studying the InMemory store for a bit, it seems to be exactly what the name suggests: it keeps stuff in memory. Which is great
If true, that means it should be pretty easy to reset its state. Just change it to whatever it was in the beginning.
That's what I attempted to do here, and it seems to work. I don't know much about the internals of commanded so I might be missing something, though

With this, instead of the whole setup described in the wiki page linked above, the test setup could just become something like:

```
setup do
  Commanded.EventStore.Adapters.InMemory.reset!()

  :ok
end
```

no need to restart/reboot anything, and also no need for `--no-start`

### DIFF
--- a/lib/commanded/event_store/adapters/in_memory.ex
+++ b/lib/commanded/event_store/adapters/in_memory.ex
@@ -106,6 +106,10 @@ defmodule Commanded.EventStore.Adapters.InMemory do
     GenServer.call(__MODULE__, {:delete_snapshot, source_uuid})
   end
 
+  def reset! do
+    GenServer.call(__MODULE__, :reset!)
+  end
+
   @impl GenServer
   def handle_call({:append_to_stream, stream_uuid, expected_version, events}, _from, %State{streams: streams} = state) do
     case Map.get(streams, stream_uuid) do
@@ -209,6 +213,10 @@ defmodule Commanded.EventStore.Adapters.InMemory do
     }
 
     {:reply, :ok, state}
+  end
+
+  def handle_call(:reset!, _from, %State{serializer: serializer}) do
+    {:reply, :ok, %State{serializer: serializer}}
   end
 
   @impl GenServer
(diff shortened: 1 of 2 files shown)


naps62:
PS: the test that is failing on travis seems to be intermittent when running locally. It also doesn't seem to be related to my changes. Can anyone clarify?

slashdotdash:
Thanks for the proposal @naps62. 

Unfortunately it's not quite as easy as providing a reset function for the in-memory event store since any event handlers and process managers also need to be reset as they have their own state (such as position of last seen event). If you reset the event store without also restarting them, then they will not behave as expected for subsequent tests.

I've experimented with resetting both the in-memory event store and handlers in the [feature/in-memory-reset](https://github.com/commanded/commanded/tree/feature/in-memory-reset) branch but it started getting more complicated than I was happy with. I also believe the dispatch consistency feature was affected due to state being held by `Commanded.Subscriptions`. 

Instead of resetting the event store, I now advise stopping your app and resetting the event store in an `on_exit` callback function defined in an ExUnit case template. Take a look at the recently updated [Testing your application](https://github.com/commanded/commanded/wiki/Testing-your-application) wiki page for details. This approach doesn't require you use the `--no-start` flag, but it does guarantee a clean environment for each test, without resulting to writing any reset code.

slashdotdash:
I've merged your change as it's a useful addition. I will update the wiki page to include its usage.

naps62:
@slashdotdash well given your first comment, I actually may disagree about it being useful

I didn't think of the issue with subscriptions. I'm relatively new to event sourcing, and only now starting to mess around here. Give what you said, won't a method like this cause issues if used improperly?

PR #183, opened 2018-07-30
Local PubSub: Replace noop Registry.match with Registry.lookup

jfornoff:
Simple change, since the `Registry.match` does not do anything in its match term, we might as well `Registry.lookup` :-)


slashdotdash:
Thanks @jfornoff.

PR #186, opened 2018-08-08
Added Additions to the wiki section to test setup

Aleksion:
@slashdotdash feel free to add the essence of this to the wiki if you prefer that. 
The solution dawned on me when I was looking through Conduit and how it's storage.ex file was setup.


slashdotdash:
Thanks for the pull request @Aleksion.

PR #196, opened 2018-09-07
Fix snapshot recording

amatalai:
## Issue

Aggregate is not properly populated from snapshots

## Reason

`snapshot_valid?` compares `aggregate.snapshot_module_version` with `snapshot.metadata["snapshot_version"]`, but when snapshot is recorded `aggregate.aggregate_version` is saved to new snapshot metadata instead of `aggregate.snapshot_module_version`. If I understand everything correctly `:aggregate_version` changes for every event while `:snapshot_module_version` should be changed by programmer inside config.exs file only when needed.

## How to prove this

* Create simple agent `Agent.start_link(fn -> 0 end, name: :foo)`
* Add `Agent.update(:foo, & (&1 + 1))` to every `apply/2` inside aggregate
* Dispatch a command
* `Agent.get(:foo, & &1)` will always return number equal to all events count

## Fix

* Changed value used to create new snapshot metadata to `:snapshot_module_version`
* Changed snapshot metadata field name to force snapshot recreation (this should be backward compatible and prevent any possible side effects)

## Things to consider

This probably require some additional tests to prevent regressions. Unfortunately, it's my first day with this library and I don't have enough time to learn its internals.

### DIFF
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -379,7 +379,7 @@ defmodule Commanded.Aggregates.Aggregate do
   defp snapshot_valid?(%SnapshotData{metadata: metadata}, %Aggregate{
          snapshot_module_version: expected_version
        }) do
-    Map.get(metadata, "snapshot_version", 1) == expected_version
+    Map.get(metadata, "snapshot_module_version", 1) == expected_version
   end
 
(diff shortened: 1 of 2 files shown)


slashdotdash:
Good catch @amatalai looks like a bug. Thank you for providing a pull request to fix my mistake.

PR #204, opened 2018-10-02
export Commanded.Commands.Router macros in .formatter.exs

amatalai:
This will allow to add `import_deps: [:commanded]` to `.formatter.exs` in projects using commanded (instead of adding every macro to locals_without_parent manually)


slashdotdash:
Thanks for the pull request @amatalai.

PR #205, opened 2018-10-04
Fixed dialyzer warning generated by "use Commanded.Commands.Router"

amatalai:
Code generated by `use Commanded.Commands.Router` was causing following dialyzer warning:
`lib/my_app/router.ex:1: Expression produces a value of type 'ok' | {'error',_}, but this value is unmatched`


slashdotdash:
Thank you!

PR #206, opened 2018-10-05
Generate specs and docs for Router dispatch functions only once.

amatalai:
At the moment dispatch macro generates spec and doc atrributes many times what makes documentation for them confusing. For example output for `iex(1)> h MyApp.Router.dispatch/1` looks like this:
screenshot 2018-10-05 at 21 58 01
Specs are accumulated and doc from register macro is completely ignored.
What's more, this also causes the following dialyzer warning:
```
lib/my_app/router.ex:4:overlapping_contract
Overloaded contract for MyApp.Router.dispatch/1 has
overlapping domains; such contracts are currently unsupported and
are simply ignored.
```

To fix this I have unified specs and moved them to place where they will be created only once: `__before_compile__` macro

### DIFF
--- a/lib/commanded/commands/router.ex
+++ b/lib/commanded/commands/router.ex
@@ -319,9 +319,6 @@ defmodule Commanded.Commands.Router do
 
       Returns `:ok` on success, or `{:error, reason}` on failure.
       """
-      @spec dispatch(command :: struct) :: :ok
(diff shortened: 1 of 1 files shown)


slashdotdash:
Good catch, thanks for the fix.

Issue #143, opened 2018-02-13
Extend aggregate lifespan behaviour to include `after_error/1`


slashdotdash:
Dispatching a command to an aggregate can return any of the following:

- One or more domain events.
- No domain events.
- An error.

To handle these scenarios the aggregate lifespan behaviour should support:

- `after_event/1`
- `after_command/1`
- `after_error/1`

The appropriate callback would be used depending upon the return value from the command dispatch.

slashdotdash:
Implemented in #210.

PR #207, opened 2018-10-09
Aggregates should shutdown after errors

uberbrodt:
Previously, the lifespan was being set to :infinity when executing a
command returned {:error, any}.


uberbrodt:
There's a case I found where this could break client code. It might make more sense to make this configurable

slashdotdash:
@uberbrodt I had previously created issue #143 to extend the aggregate lifespan behaviour to include an `after_error/1` callback. That approach would allow consumers to decide whether to stop the aggregate on error.

uberbrodt:
I think I can turn around an implementation for #143 today, that will be better in the long run.

uberbrodt:
Closing this as it's superseded by #210

PR #210, opened 2018-10-11
Expands AggregateLifespan to include after_command/1 and after_error/1 

uberbrodt:
Previously, aggregates would not shut down if one or more events were
not returned after command execution. This change defines callbacks to
handle cases where an error or a [] is returned.


slashdotdash:
Thanks for taking the time to implement #143 in full, much appreciated.

slashdotdash:
Closes #143.

PR #216, opened 2018-10-26
Fix typo in router option (fixed #208)

### LINKED ISSUE #208: Global option `:include_execution_result` not taking effect
I went into the code and saw its just typo that was causing the problem

https://github.com/commanded/commanded/blob/c8cd25527f327679ac76928129dc368bc6f92b7e/lib/commanded/commands/router.ex#L149

`:incldue_execution_result` has a typo, it should be `:include_execution_result`

janpieper:
I don't think this PR needs a description :wink:


slashdotdash:
Thank you @janpieper . 

Closes #208.

PR #222, opened 2018-11-11
Fix subscribe/1 '@spec'

joaothallis:
There are places where `subscribe/1` `@ spec` is expecting string and not just atom


slashdotdash:
Thanks for fixing this issue.

PR #223, opened 2018-11-11
Change function to private

joaothallis:
name/1 function is used only in handler.ex


slashdotdash:
Thank you @joaothallis.

PR #224, opened 2018-11-11
Remove pattern that can not occur

joaothallis:
`Task.yield/1` and `Task.shutdown/1` do not return: `{:error, error}`, only `{:ok, term} | {:exit, term} | nil`


slashdotdash:
Thanks for the fix @joaothallis.

PR #230, opened 2018-11-25
Remove unnecessary else

joaothallis:
With the error being returned without any changes, you do not need the else, only if you want to be explicit


slashdotdash:
Good catch, I'd forgotten that's the default behaviour of the `with` clause.

PR #234, opened 2018-12-27
Moving to Jason

BlueHotDog:
Hi,
This is still work in progress but wanted to share to get some feedback.
As you know Ecto moved away to Jason in v3, which breaks Commanded.
This PR is an effort to align Commanded with Ecto by moving it to Jason as well.
I've made a couple assumptions in this PR which i'm happy to discuss:
- It's a breaking change. Since Jason requires explicit ```@derive Jason.Encoder``` vs the implicit behavior in Poison.. If you think this is too harsh of a change, i think we can, with more work, do something to have a better upgrade experience. E.g have a Commanded.JsonEncoder which will define the encoding behaviour depending if Jason or Poison is loaded and basically delegate to the appropriate ```@derive```. But this is quite a lot of work and i'm not sure its worth the effort - What do you think?
- Aside from that the only change is in how Jason handles NaiveDateTime, seems like Poison automatically decodes it to DateTime when Jason keeps it as string for manual decoding. Not sure how to overcome this or whats the best approach - Suggestions are welcome(1 test is failing due to this)

I've updated the documentation but i think more needs to be written in regard to this change.

Help appreciated.
Thank you for the awesome lib!

### DIFF
--- a/guides/Events.md
+++ b/guides/Events.md
@@ -8,6 +8,7 @@ Create a module per domain event and define the fields with `defstruct`. An even
 
 ```elixir
 defmodule BankAccountOpened do
+  @derive Jason.Encoder
   defstruct [:account_number, :initial_balance]
 end
 ```
--- a/guides/Getting Started.md
+++ b/guides/Getting Started.md
@@ -10,11 +10,11 @@ Commanded can be installed from hex as follows.
     end
     ```
 
-2. Optionally add `poison` to make `Commanded.Serialization.JsonSerializer` available:
+2. Optionally add `jason` to make `Commanded.Serialization.JsonSerializer` available:
 
     ```elixir
     def deps do
-      [{:poison, "~> 3.1 or ~> 4.0"}]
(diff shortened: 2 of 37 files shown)


BlueHotDog:
The commit messages are a mess, i'll make sure to rewrite them before merge.

slashdotdash:
@BlueHotDog Thanks for taking the time to work on this pull request, it's really helpful. 

I'm finally looking at finishing off the Jason upgrade for Event Store and Commanded so will take this and get it merged in. 

I think it's acceptable to have a breaking change with Jason requiring explicit `@derive Jason.Encoder` in event modules for serialization. As for date time parsing we can support it via the existing `Commanded.Serialization.JsonDecoder` protocol.

slashdotdash:
@BlueHotDog I've rebased your changes on master and have merged in 6dcc60b11e5af897167c72836c3e3e58f1ce204c.

BlueHotDog:
Whoo hoo!! excited!

PR #246, opened 2019-01-23
Update typespec for EventData.

m-doughty:
`data` should be a struct, not a binary.
`metadata` should be a map, not a binary.


slashdotdash:
Thanks for fixing this @m-doughty.

Issue #169, opened 2018-04-18
Filter specific fields from logged command


slashdotdash:
Commanded logs commands during execution (command dispatch) using the debug level. To prevent sensitive data from being exposed in the log files the command fields should be filtered.

Consumers can use Elixir's built-in `Inspect` module to exclude certain fields from being logged, as shown below.

```elixir
defmodule RegisterUser do
  @derive {Inspect, only: [:id, :name]}
  defstruct [:id, :name, :password]
end

inspect(%RegisterUser{id: 1, name: "Homer", password: "D'oh"})
"#RegisterUser<id: 1, name: \"Homer\", ...>"
```

The above advice should be included in the command guide.

PR #249, opened 2019-01-28
Add filter configuration for commanded log

Freyskeyd:
Hello,

I've implemented something for #169 

Dunno if still needed, I will try to add some tests if necessary.

Signed-off-by: Freyskeyd <simon.paitrault@gmail.com>

### DIFF
--- a/lib/commanded/commands/decriber.ex
+++ b/lib/commanded/commands/decriber.ex
@@ -1,11 +1,18 @@
 defmodule Commanded.Commands.Describer do
-  @moduledoc false
+  @moduledoc """
+  Commands describer is used to describe a command, mostly for filtering sensitive data from being exposed in log outputs.
 
-  @fields_to_filter Application.get_env(:commanded, :filter_fields, [])
+  ## Example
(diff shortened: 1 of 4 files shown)


lib/commanded/commands/decriber.ex
```diff
@@ -0,0 +1,26 @@
+defmodule Commanded.Commands.Describer do
+  @moduledoc """
+  Commands describer is used to describe a command, mostly for filtering sensitive data from being exposed in log outputs.
+
+  ## Example
+      config :commanded, :filter_fields, [:password, :secret_key]
+  """
+
+  @fields_to_filter Application.get_env(:commanded, :filter_fields, [])
```
slashdotdash: The `Application.get_env(:commanded, :filter_fields, [])` call will need to be done at runtime, not compile time, or else changing the config will not have an effect until commanded dependency is recompiled (e.g. `mix deps.compile --force command`).

Freyskeyd:
@slashdotdash I fixed the problem and added some tests.

Freyskeyd:
you're right!

PR #251, opened 2019-02-03
Remove `account` var in guides/Usage.md, it's not used

kpanic:


slashdotdash:
Thanks @kpanic.

kpanic:
Thanks to you @slashdotdash for commanded!

PR #262, opened 2019-03-16
Add include_execution_result and aggregate_version to type spec.

trbngr:
This fixes dialyzer warnings when pattern matching on the results of `Router.dispatch(cmd, include_execution_results: true)`


slashdotdash:
Thanks for the pull request Chris.

Issue #258, opened 2019-03-05
Event upcasting


slashdotdash:
Add support to Commanded for runtime event upcasting. 

Upcasting allows an event to be transformed before being passed to a consumer. You can use an upcaster to change the shape of an event (e.g. add a new field with a default, rename a field). Because the upcaster changes any historical event to the latest version, consumers (aggregates, event handlers, and process managers) only need to support the last version.

The upcaster could be implemented as an Elixir protocol, similar to the existing `Commanded.Serialization.JsonDecoder` protocol. It must have access to the raw event data and metadata, including event type.

trbngr:
I've issued a PR of a spike of this feature.

An interesting feature of this is the ability to "adapt" an old event into a completely event type. This is a pretty powerful way of dealing with "legacy" event streams with newer code.

trbngr:
This has been merged. Thanks Ben!!

slashdotdash:
Done in #263.

PR #263, opened 2019-03-16
Implement event upcasting.

trbngr:
A quick stab at #258

### DIFF
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -423,10 +423,14 @@ defmodule Commanded.Event.Handler do
 
   @doc false
   def handle_info({:events, events}, %Handler{} = state) do
+    alias Commanded.Event.Upcasting
     Logger.debug(fn -> describe(state) <> " received events: #{inspect(events)}" end)
 
     try do
-      state = Enum.reduce(events, state, &handle_event/2)
+      state =
+        events
+        |> Upcasting.upcast_event_stream()
+        |> Enum.reduce(state, &handle_event/2)
 
       {:noreply, state}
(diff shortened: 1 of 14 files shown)


slashdotdash:
Thanks for the pull request, this looks like a good start. 

It’s worth noting that events published to subscribers, such as event handlers and process managers, won’t go through the stream forward function. To solve this issue the `Commanded.EventStore` behaviour will need to be extended to allow a mapper function to be provided to both stream forward and subscribe to stream so that the upcasting can happen in both places.

trbngr:
Thanks Ben. Good catch. I'd like to investigate this further before you merge it.

trbngr:
I think the latest commit addresses your concerns and will not required tweaking on the ES adapters.

As I believe this is `commanded` only feature, we should be good.

trbngr:
Oh! And I fixed a bug in the `InMemory` adapter. It was enuming over the stream and therefore executing the event_stream.

trbngr:
I wonder if `stream_forward` is the right place for the upcast. I'm thinking it should really be in `Aggregate.handle_info({:events, events}, _)` like the others.

Thoughts?

slashdotdash:
> Aggregate.handle_info({:events, events}, _)

Is only used for transient subscriptions to the configured event store which means it will never receive historical events so no upcasting is required.

Having the upcast applied in `stream_forward` is useful because it allows consumers to read from the configured event store if they need to.

slashdotdash:
@trbngr Is this PR ready to merge now?

trbngr:
That's your call, bud. I am done with it

PR #273, opened 2019-05-09
Modify Registration.start_child/3 to pass a child_spec

uberbrodt:
In order to support DynamicSupervisor, Commanded.Registration.start_child/3
needed to receive a child_spec instead of args like a :simple_one_for_one
supervisor would expect. Supporting DynamicSupervisor will allow us to
support distributed supervisors like Horde.Supervisor.


slashdotdash:
Thanks for the pull request Chris.

PR #274, opened 2019-05-10
Add a config option to define the Aggregates.Supervisor :start function

uberbrodt:
This will be used by [Commanded.Registration.HordeRegistry](https://github.com/uberbrodt/commanded_horde_registry) to start a `Horde.Supervisor` instead of a `DynamicSupervisor`. I'm definitely open to alternatives, this seemed like the least amount of work to accomplish the goal.

### DIFF
--- a/lib/commanded/aggregates/supervisor.ex
+++ b/lib/commanded/aggregates/supervisor.ex
@@ -3,12 +3,20 @@ defmodule Commanded.Aggregates.Supervisor do
   Supervises `Commanded.Aggregates.Aggregate` instance processes.
   """
 
-  use DynamicSupervisor
-
   require Logger
 
   alias Commanded.Aggregates.Aggregate
   alias Commanded.Registration
+
+  def child_spec(arg) do
+    default = %{
+      id: __MODULE__,
+      start: Application.get_env(:commanded, :aggregate_supervisor_mfa, {__MODULE__, :start_link, [arg]}),
+      type: :supervisor
+    }
+
+    Supervisor.child_spec(default, [])
+  end
 
   def start_link(arg) do
     DynamicSupervisor.start_link(__MODULE__, arg, name: __MODULE__)


edwardzhou:
nice, I am writing horde registry plugin like commanded-swarm-registry.

slashdotdash:
@uberbrodt Instead of requiring the user to configure the `aggregate_supervisor_mfa` in config, it would be easier if we extended the `Commanded.Registration` behaviour with another callback function to allow the configured adapter to provide the supervisor spec. This reduces the configuration burden on the end user to just defining the `registry` module config option.

```elixir
defmodule Commanded.Registration do
  @callback supervisor_child_spec(module :: atom, arg :: any()) :: :supervisor.child_spec()

  @spec  supervisor_child_spec(module :: atom, arg :: any()) :: :supervisor.child_spec()
  def supervisor_child_spec(module, arg), do: registry_provider().supervisor_child_spec(module, arg)
end
```

The default local registry could define the new `supervisor_child_spec/2` function as:

```elixir
defmodule Commanded.Registration.LocalRegistry do
  @spec  supervisor_child_spec(module :: atom, arg :: any()) :: :supervisor.child_spec()
  def supervisor_child_spec(module, arg) do
    default = %{
     id: module,
     start: {module, :start_link, [arg]},
     type: :supervisor
    }

    Supervisor.child_spec(default, [])
  end
end
```

Finally, the aggregate supervisor could use the new `Registration.supervisor_child_spec/2` function to define its own child spec:

```elixir
defmodule Commanded.Aggregates.Supervisor do
  def child_spec(arg) do
    Registration.supervisor_child_spec(__MODULE__, arg)
  end
end
```

Your Horde registry could define the above function as required for Horde supervision.

What do you think of this approach?

uberbrodt:
I think that could work. I'll try it out.

uberbrodt:
So your proposal @slashdotdash works, but to work with Horde we still need a way to pass some options since `Commanded.Aggregates.Supervisor` is started as part of the `:commanded` supervision tree. We could do this a couple of ways:

1. Just have a Horde specific config in `Commanded.Registration.HordeRegistry` like so:
```elixir
  @impl Commanded.Registration
  def supervisor_child_spec(_module, _args) do
    defaults = [
      strategy: :one_for_one,
      distribution_strategy: Horde.UniformDistribution,
      name: Commanded.Aggregates.Supervisor
    ]

    opts = Application.get_env(:commanded, :horde_supervisor_opts, defaults)

    Horde.Supervisor.child_spec(opts)
  end
```
2. Alternatively, we could pass them from the `Commanded.Supervisor` via a general config option:
```elixir
  def init(_) do
    aggregate_supervisor_opts = Application.get_env(:commanded, :aggregate_supervisor_opts, [])
    children =
      EventStore.child_spec() ++
        Registration.child_spec() ++
        PubSub.child_spec() ++
        [
          {Task.Supervisor, name: Commanded.Commands.TaskDispatcher},
          {Commanded.Aggregates.Supervisor, aggregate_supervisor_opts},
          Commanded.Subscriptions.Registry,
          Commanded.Subscriptions
        ]

    Supervisor.init(children, strategy: :one_for_one)
  end
```

Option 2 seems a bit more general (and passes config as parameters, which seems to be the current "Best Practice" regarding config), but Option 1 will work fine for my purposes I think.

slashdotdash:
@uberbrodt Is it just the `distribution_strategy` for Horde that needs to be user configurable? That would simplify the configuration to just be Horde specific (a variant of option 1 above).

uberbrodt:
@slashdotdash We could probably get away with just a config option for `:distribution_strategy`. Looks like the other options are all standard `DynamicSupervisor` init options.

slashdotdash:
@uberbrodt That would be good, can you update the pull request.

uberbrodt:
Looks like I ran into an issue with Horde (or my code!). Once I get that resolved and I'm sure I don't need any other changes in Commanded I'll update the PR.

uberbrodt:
I ended up just doing a separate PR #277 . Closing this one.

PR #277, opened 2019-05-14
Add supervisor_child_spec/2 to Commanded.Registration

uberbrodt:
This will allow us to swap out Supervisor implementations, supporting
Horde.Supervisor for example. Supersedes the approach in #274


slashdotdash:
Thanks Chris.

PR #287, opened 2019-06-05
Fix Aggregate initialization races

uberbrodt:
It's possible for an Aggregate to receive an command before it's state
is initialized. This leads to execute/2 handlers receiving `nil` for the
aggregate state. It appears to happen more often when using a
distributed process registry. The fix is to replace the casts in the
init/1 handler with a chain of handle_continue/2 callbacks that are
guaranteed to complete before the first message is handled.

COMPATIBILITY NOTE: GenServer.handle_continue/2 requires OTP-21 or greater


slashdotdash:
Thanks for the pull request @uberbrodt.

PR #290, opened 2019-06-13
Process manager idle process timeout

### LINKED ISSUE #148: Process manager idle timeout
Allow an idle timeout to be specified for process manager processes.

Two possible approaches:

1. Define an `idle_timeout`, in milliseconds, as an option to start link allowing it to be configured when including a process manager in a supervisor:

```elixir
children = [
  {MyApp.FooManager, idle_timeout: 60_000},
]
```

2. Define a lifespan behaviour, similar to aggregate lifespan, to be defined per process manager. It would decide when to shutdown the process depending upon the recently dispatched command, handled event, or error.

slashdotdash:
Every started instance of a process manager runs indefinitely. To reduce memory usage this pull request adds an idle timeout configuration (in milliseconds) after which an instance process will be shutdown.

The process will be restarted on demand, whenever an event is routed to it, and its state will be rehydrated from the instance snapshot.

## Example

Define the `idle_timeout` option in the process manager module:

```elixir
defmodule ExampleProcessManager do
  use Commanded.ProcessManagers.ProcessManager,
    name: "ExampleProcessManager",
    router: ExampleRouter,
    idle_timeout: :timer.minutes(10)
end
```

Or provide the timeout on start:

```elixir
ExampleProcessManager.start_link(idle_timeout: 10_000)
```

By default the idle timeout will be configured as `:infinity` so every process manager instance will run indefinitely once started.

Fixes #148.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: elixir
 
 elixir:
-  - 1.8.0
+  - 1.8.2
 
 otp_release:
   - 21.2
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -290,11 +290,11 @@ defmodule Commanded.Event.Handler do
   def parse_name(_module, name), do: inspect(name)
 
   @doc false
-  def start_opts(module, module_opts, local_opts) do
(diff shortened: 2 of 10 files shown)


slashdotdash:
Published to Hex as v0.18.1 and v0.17.5 releases.

PR #291, opened 2019-06-26
Support {:system, varname} format in Phoenix PubSub

justmendes:
This is more of a PR for discussion on the best approach.

Currently the supervisor is calling `PubSub.child_spec/0` which in turn sets passes the options defined in `config/*.exs` down to `Phoenix.PubSub`.

With this setup, it is non-trivial to set values in runtime (like some environment variables which are not always available in compile time) for PubSub. It would be possible to overwrite the supervisor but we would need to explicitly start the tracker, among others.

The approach we went with was to change the `Commanded.PubSub.PhoenixPubSub` module to account for variables in the format of `{:system, varname}`. This is very hackish, however. Ideally this could be something set in the supervisor and propagated down or even to use an existing PubSub instance started in the supervisor, instead of spawning one.

/cc @naps62

### DIFF
--- a/lib/commanded/pubsub/phoenix_pubsub.ex
+++ b/lib/commanded/pubsub/phoenix_pubsub.ex
@@ -66,6 +66,7 @@ if Code.ensure_loaded?(Phoenix.PubSub) do
         |> Keyword.put(:name, __MODULE__)
 
       {adapter, phoenix_pubsub_config} = Keyword.pop(phoenix_pubsub_config, :adapter)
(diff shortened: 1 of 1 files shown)


slashdotdash:
Thanks for the pull request @justmendes. 

This approach will work for now as I plan to migrate Commanded from a singleton application to allow consumers to define their own app specific instances which will require config to be provided top-down.

justmendes:
@slashdotdash that would be perfect! Thanks for merging the momentary fix! 👍

PR #298, opened 2019-08-01
Support multiple Commanded apps

### LINKED ISSUE #117: Concurrent, Isolated commanded apps
*This is a very hard thing to implement, but lets discuss it*
*Also, I wrote this while rushing a little bit, I'll refine it later*

# Problem

This is somehow related to #113, but this is a little bit more.
I think that this approach will solve #113 and some other usecases.

One of the biggest problems of commanded right now (to me at least) is the "global" nature of it. Right now we configure an app using configs and that means that we can have only one instance per app (including umbrella apps).

This also makes testing very hard since we cannot run tests concurrently and we have to always reset the event store to ensure tests doesn't interfere with one another.

Another thing, as pointed by #113 is that we may want to have different EventStores per app.

Right now, the only way of having "different applications" is to use multiple routers, however, they all use the same eventstore, then the applications are not isolated (although it may be a good idea to have them connected through the same eventstore so we can have shared subscriptions to allow event-collaboration workflows, otherwise people would need some kind of integration middleware to propagate events)

# Goals

Summing up, my goals are (in no specific order):
- Allow isolated instances to be started at test (create and destroy applications on the go)
- Allow multiple instances to run for different apps in the same umbrella/cluster
- Be backwards compatible (as much as it is possible) with our current way of doing things
- Start the applications via configuration by default (something like Ecto does with the repos, by specifying an otp_app)
- Allow dispatching to a specific instance and directly using module (using a registration as it is right now)

# Simplified overview

Just a fast overview so we know where we need to make changes

@slashdotdash please correct me where I'm wrong, which is probably in a lot of places since this is my fast overview over the code

A commanded application can be said to be composed by (and some other pieces):
- Router (runs on caller process)
- Dispatcher (runs on caller process)
- Pipeline/Middleware Layer (runs on caller process)
- Aggregate Supervisor
- Aggregate processes
- Event Store
- Event Handlers
- Registration
- Subscriptions

The processes involved are generaly

1. A router maps given commands to aggregates and then asks the dispatcher to dispatch it (via dispatcher module)
2. The dispatcher
    1. Executes the pipeline
    2. Opens the aggregate (through the aggregate supervisor, which starts a new process and rehydrates with events from the eventstore or just returns the pid)
    3. Starts a supervised task that calls the aggregate genserver (via name: Aggregate Name + Aggregate uuid)
3. The aggregate process
    1. Given the execution context, runs the handle function in the handler module or the execute on the aggregate module
    2. Tries to persist the events (calling EventStore directly)
4. EventStore work (let's not get into this here)
    1. In background and async, the event handlers will catch up with event_store
5. Dispatcher executes the pipeline according to the result
    1. There is one middleware that if `consistency: :strong` is passed, it waits on all event handlers using `Subscriptions.wait_for`

# Proposal

Right now, the entrypoint for the application is the router.
I suggest including a new concept: An Application.

```elixir
defmodule MyApplication
  use Commanded,
    router: MyRouter,
    event_store: MyEventStore
end
```

There would be three modus operandis:

**Complete isolation**
```elixir
my_app = MyApp.start_link
MyApp.dispatch(cmd, application: my_app)
```
**Per Application**
```elixir
MyApp.dispatch(cmd)
```

**As it is**

```elixir
MyRouter.dispatch(cmd)
```

The way it works is as follows:
We'll introduce a new option for dispatch, which it is the pid of the application 

`MyRouter.dispatch(cmd, application: pid)`
If the pid is not found, we'll lookup for a global application (which right now is started by :commanded otp app)

MyApp.dispatch is the same thing as the router, however, it looks up for the global application scoping by application module name (MyApp, MyOtherApp, etc) but also accepts the myapp pid.
It will need to lookup the router as well (which was given via the `use Commanded` opts)

# Challanges

There a lot of places in the codebase where we use global registries. We may need to allow scoping these registries using Module names and pids. This includes:
- `Aggregates.Supervisor.open_aggregate`
- `EventStore.*`
- `Subscriptions.*`

There is also a require a change on how the commanded application is started.

# Inspiration

We may want to have a look at how `Ecto` does that

slashdotdash:
Add support for defining one or more Commanded application modules for your Elixir app. 

Each application is independent, isolated, and its lifecycle is controlled by you (started in your app's own supervision tree).

## Examples

A basic application defines the `otp_app` (for its configuration) and includes one or more router modules.

```elixir
  use Commanded.Application, otp_app: :my_app

  router MyApp.Router
end
```

The above application could be configured with:

```elixir
# config/config.exs
config :my_app, MyApp.Application
  event_store: [
    adapter: Commanded.EventStore.Adapters.EventStore,
    event_store: MyApp.EventStore
  ],
  pub_sub: :local,
  registry: :local
```

Alternatively, you can include the event store, pubsub, and registry config when defining the application:

```elixir
  use Commanded.Application,
    otp_app: :my_app,
    event_store: [
      adapter: Commanded.EventStore.Adapters.EventStore,
      event_store: MyApp.EventStore
    ],
    pubsub: :local,
    registry: :local

  router(MyApp.Router)
end
```

Commands are dispatched using the Application:

```elixir
:ok = MyApp.Application.dispatch(%ExampleCommand{})
```

For backwards compatibility commands can be dispatch using a router but must specify the application as an option:

```elixir
:ok = MyApp.Router.dispatch(%ExampleCommand{}, application: MyApp.Application)
```

Fixes #117.


andreapavoni:
Congrats! this is one of the key features I really needed.

Issue #279, opened 2019-05-16
Exception when command dispatched immediately after aggregate process shutdown


slashdotdash:
Dispatched commands are sent to an aggregate `GenServer` process using the Elixir `Registry` module (by default). When using an [aggregate lifespan](https://github.com/commanded/commanded/blob/master/guides/Commands.md#aggregate-lifespan) you can choose to stop the aggregate process after an event, command, or error.

```elixir
defmodule AccountLifespan do
  @behaviour Commanded.Aggregates.AggregateLifespan

  def after_event(%AccountDeleted{}), do: :stop 
end
```

If another command is sent to the same aggregate process immediately after it has been stopped the `Registry` may not have removed the stopped PID and so the command can be sent to the terminated process.

```
** (stop) exited in: GenServer.call({:via, Registry, {Commanded.Registration.LocalRegistry, {Account, "1"}}}, {:execute_command, %Commanded.Aggregates.ExecutionContext{causation_id: "54b91ecc-92a4-46b3-b06d-ec75895cd64d", command: %NotifyAccountDisconnected{account_id: "1"}, correlation_id: "9360db25-a448-4a1e-a8ac-87f9c656b707", function: :handle, handler: AccountHandler, lifespan: AccountLifespan, metadata: %{}, retry_attempts: 10}}, 5000)
    ** (EXIT) normal
    (elixir) lib/gen_server.ex:989: GenServer.call/3
    (elixir) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
    (elixir) lib/task/supervised.ex:35: Task.Supervised.reply/5
    (stdlib) proc_lib.erl:249: :proc_lib.init_p_do_apply/3
Function: &Commanded.Aggregates.Aggregate.execute/4
    Args: [Account, "1", %Commanded.Aggregates.ExecutionContext{causation_id: "54b91ecc-92a4-46b3-b06d-ec75895cd64d", command: %NotifyAccountDisconnected{account_id: "1"}, correlation_id: "9360db25-a448-4a1e-a8ac-87f9c656b707", function: :handle, handler: AccountHandler, lifespan: AccountLifespan, metadata: %{}, retry_attempts: 10}, 5000]
```

This exception should be handled by starting the aggregate process and sending it the command being dispatched.

The [commanded_aggregate_race](https://github.com/Freyskeyd/commanded_aggregate_race) repository reproduces this issue and should be used to verify any fix.

slashdotdash:
Fixed by #308.

PR #308, opened 2019-10-27
Fix commanded aggregate race

fahchen:
address https://github.com/commanded/commanded/issues/279


fahchen:
@slashdotdash Could you have a look, please?

slashdotdash:
Thanks for the pull request @fahchen. Do you think there is any possibility the dispatcher could enter an infinite loop when attempting to execute the command dispatch after the aggregate process exits, and then repeatedly do so?

fahchen:
The exit is caught by pattern matching, so we only retry when the process exits for lifespan(the exit reason is normal actually). If the aggregate is exited with normal reason  for something other than lifespan, we can’t distinguish from lifespan.
So I assume that the lifespan is the only reason when the process exits, we don’t catch and retry any other reasons.

Issue #299, opened 2019-08-07
Allow serialization options to be configured for the JSON serializer


slashdotdash:
Jason allows the following options to be provided when decoding JSON strings:

- `:keys` - controls how keys in objects are decoded. Possible values are:
  - `:strings` (default) - decodes keys as binary strings
  - `:atoms` - keys are converted to atoms using `String.to_atom/1`
  - `:atoms!` - keys are converted to atoms using `String.to_existing_atom/1`
- `:strings` - controls how strings (including keys) are decoded. Possible values are:
  - `:reference` (default) - when possible tries to create a sub-binary into the original
  - `:copy` - always copies the strings. This option is especially useful when parts of the decoded data will be stored for a long time (in ets or some process) to avoid keeping the reference to the original data.

It should be possible to configure these options in Commanded to be passed to the configured JSON serializer.

APiercey:
I would like to try and pick this up :)! I'll see what I can do and put up a PR.

APiercey:
PR here: https://github.com/commanded/commanded/pull/309

PR #309, opened 2019-10-28
Feature: Allow Jason serialization options

APiercey:
From: https://github.com/commanded/commanded/issues/299

This allows encoding and decoding options to be passed to the serializer.

If it works the project, I can use the `SerializationCase` to implement the functionality in other adapters. As I understand, these two are the supported ones:
https://github.com/commanded/commanded-extreme-adapter
https://github.com/commanded/commanded-eventstore-adapter


slashdotdash:
Thanks for the pull request @APiercey.

Due to the way that the EventStore uses the configured serializer it would also need to be changed to support passing the encode and decode options.

We could extend the `serializer` config option to support an optional tuple format which includes both the serializer module and the serialization options:

```elixir
serializer: {Serializer, encode: [..], decode: [..]}
```

Using the in-memory event store with Commanded:

```elixir
  use Commanded.Application,
    otp_app: :my_app,
    event_store: [
      adapter: Commanded.EventStore.Adapters.InMemory,
      serializer:
        {Commanded.Serialization.JsonSerializer,
         encode: [escape: :json], decode: [keys: :strings]}
    ]
end
```

Using a Postgres EventStore with Commanded:

```elixir
  use Commanded.Application,
    otp_app: :my_app,
    event_store: [
      adapter: Commanded.EventStore.Adapters.EventStore,
      event_store: MyApp.EventStore
    ]
end
```

```elixir
# config/config.exs
config :my_app, MyApp.EventStore,
  username: "postgres",
  password: "postgres",
  database: "eventstore",
  hostname: "localhost",
  serializer:
    {Commanded.Serialization.JsonSerializer,
     encode: [escape: :json], decode: [keys: :strings]}
```

It should still be possible to configure the serializer without providing any options as:

```elixir
serializer: Commanded.Serialization.JsonSerializer
```

What are your thoughts?

lib/commanded/serialization/json_serializer.ex
```diff
@@ -10,28 +10,45 @@ if Code.ensure_loaded?(Jason) do
...
     Deserialize given JSON binary data to the expected type.
     """
     def deserialize(binary, config \\ []) do
-      {type, opts} =
-        case Keyword.get(config, :type) do
-          nil -> {nil, %{}}
-          type -> {TypeProvider.to_struct(type), %{keys: :atoms}}
+      type = maybe_type(config)
+
+      options =
```
joaothallis: What do you think about use `if` instead of `case` for cases in what is binary?
```suggestion
      options = if type, do? Map.merge(%{keys: :atoms}, maybe_options(config)), else: %{}
```

jsmestad:
I am hitting this issue with a nested map trying to cast to existing atom.

@slashdotdash I think the ability to configure the serializer may be better on a specific serializer than globally?

Issue #303, opened 2019-09-16
Include stacktrace in event handler and process manager `error` callback functions


slashdotdash:
To assist with debugging exceptions the stacktrace should be included in the failure context struct passed to the event handler and process manager error callback functions. This would allow the user to log the error and the source stacktrace.

justgage:
This has been a pretty big pain point, I think I might be one I take on. Basically any time that an event handler crashes it causes a huge error and usually the source of the problem is near the beginning.

davydog187:
@justgage @slashdotdash I pushed up a PR here. Please take a look

davydog187:
I'm happy to add another to process manager. Note that I did not pass to the callback as it didn't seem necessary.

davydog187:
#340 is merged :)

davydog187:
Im going to work on a PR for process managers as well. Afterwards a release of commanded would be awesome!

We’re about to put Commanded into production at SimpleBet, and this will be quite helpful for debugging 😅

slashdotdash:
@davydog187 I've started work on adding `__STACKTRACE__` to event handler and process manager failure contexts in branch [ent-handler-error-stacktrace](https://github.com/commanded/commanded/tree/feature/event-handler-error-stacktrace). Process managers are slightly more complex than event handlers because they can error when handling an event or when dispatching a command.

davydog187:
@slashdotdash great! I didn't start it yet, so if you've got it I won't bother Happy to help though!

davydog187:
Took a look at your branch. You know the codebase a million times better than I do. It looks great!

PR #340, opened 2020-02-16
Add stacktraces to EventHandler error logging

davydog187:
### Issue
#303 

### Description
This PR logs the stacktrace when a user-written EventHandler raises an error. 

Note that I hacked in a test scenario, there might be a better way of doing that. I'm open to suggestions.

Screen Shot 2020-02-16 at 9 35 45 AM

### DIFF
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -551,7 +551,7 @@ defmodule Commanded.Event.Handler do
   @doc false
   @impl GenServer
   def handle_info(message, state) do
-    Logger.error(fn -> describe(state) <> " received unexpected message: " <> inspect(message) end)
+    Logger.error(fn -> describe(state) <> " received unexpected message: " <> inspect(message, pretty: true) end)
 
     {:noreply, state}
   end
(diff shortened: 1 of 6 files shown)


lib/commanded/event/handler.ex
```diff
@@ -626,6 +626,15 @@ defmodule Commanded.Event.Handler do
         end)
 
         handle_event_error(error, event, state, context)
+
+      {:error, reason, stacktrace} ->
+        Logger.error(fn ->
+          describe(state) <>
+            " failed to handle event #{inspect(event)} due to: #{inspect(reason)}"
```
davydog187: In the future, I would suggest we move to the Exception formatter to log these. I can open another issue if we think that's a good idea.

lib/commanded/event/handler.ex
```diff
@@ -626,6 +626,15 @@ defmodule Commanded.Event.Handler do
...
+        end)
+        Logger.error(Exception.format(:error, reason, stacktrace))
```
davydog187: I should use the function style logs

lib/commanded/event/handler.ex
davydog187: I added the `pretty` option to inspect the error messages. It makes the log much easier to read.

Screen Shot 2020-02-17 at 7 26 21 AM

slashdotdash:
Thanks Dave.

PR #341, opened 2020-02-16
Run the formatter in CI

davydog187:
This PR makes sure that all code pushed to `commanded` is formatted. I'm initially pushing it up unformatted to show that it works, then I will format and push a second commit.


slashdotdash:
That's really useful, thanks Dave.

PR #346, opened 2020-02-19
Add the supervisor_child_spec back to the Aggregates supervisor

oestrich:
This was deleted somewhere along the way and the horde registry
requires this.

### DIFF
--- a/lib/commanded/aggregates/supervisor.ex
+++ b/lib/commanded/aggregates/supervisor.ex
@@ -10,6 +10,11 @@ defmodule Commanded.Aggregates.Supervisor do
   alias Commanded.Aggregates.Aggregate
   alias Commanded.Registration
 
+  def child_spec(arg) do
+    application = Keyword.fetch!(arg, :application)
+    Registration.supervisor_child_spec(application, __MODULE__, arg)
+  end
+
   def start_link(args) do
     application = Keyword.fetch!(args, :application)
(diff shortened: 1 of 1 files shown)


slashdotdash:
@oestrich This function was removed because it was unused by Commanded. I can't see where it is used by the Horde registry, could you show me?

oestrich:
It spins up a horde dynamic supervisor which manages distributing processes across the cluster. I'm not 100% on the integration works as I'm attempting to upgrade the commanded horde registry for 1.0 and ran into this.

https://github.com/uberbrodt/commanded_horde_registry/blob/master/lib/commanded/registration/horde_registry.ex#L32

Alternatively, if it's not used, we should remove the callbacks for it, it took a while to figure out that the callback existed but was unused.

uberbrodt:
Yes it does look we'll need this back. It's used in `Commanded.Aggregates.Supervisor`, in place of the call to `DynamicSupervisor.start_link/3`. 

@oestrich you may want to work off the `testing` branch of `commanded_horde_registry` which is where I'm testing integration with Commanded 1.0

uberbrodt:
I've published an [alpha version](https://hex.pm/packages/commanded_horde_registry/1.0.0-alpha.0) of `commanded_horde_registry` that doesn't use a Horde.DynamicSupervisor in start_child/3. Seems to work in my testing.

Issue #353, opened 2020-03-10
Wrong aggregate_version after rebuild from events


CptBreeza:
## Background
After a refactor we decided to merge two streams that were used by two aggregates to one new stream which would be used by some other aggregate. We linked the events of the two original streams to a new stream when migrating.

## The Problem
An aggregate rebuilds its state by events, of which the `stream_version` fields are filled by the `original_stream_version` column(in the case of using [commanded/eventstore](https://github.com/commanded/eventstore)).

Then the aggregate append events with an expected_version which is the last `stream_version` of the events that it used to rebuild its state. But the `append_to_stream` checks the expected_version with the `stream_version` taken from `streams` table.

The problems is that the last `original_stream_version` value that is taken from `stream_events` table could be different from the `stream_version` value that is taken from `streams` table when operating on a stream of which the events are linked from some different streams.

## Example
Suppose there's a stream with stream_id = 1 and stream_events
| event_id | stream_id | stream_version | original_stream_id | orignal_stream_version |
|---|---|---|---|---|
| a | 1 | 1 | 1 | 1 |
| b | 1 | 2 | 1 | 2 |

and a stream with stream_id = 2 and stream_events
| event_id | stream_id | stream_version | original_stream_id | orignal_stream_version |
|---|---|---|---|---|
| c | 2 | 1 | 2 | 1 |
| d | 2 | 2 | 2 | 2 |
| e | 2 | 3 | 2 | 3 |

After linking events from those two streams by some order to a new stream
| event_id | stream_id | stream_version | original_stream_id | orignal_stream_version |
|---|---|---|---|---|
| c | 3 | 1 | 2 | 1 |
| a | 3 | 2 | 1 | 1 |
| b | 3 | 3 | 1 | 2 |
| d | 3 | 4 | 2 | 2 |
| e | 3 | 5 | 2 | 3 |

And the `streams` table would be like
| stream_id | stream_uuid | stream_version |
|---|---|---|
| 1 | 1 | 2 |
| 2 | 2 | 3 |
| 3 | 3 | 5 |

An aggregate would get an aggregate_version of **3** after rebuilding state from stream_3, but the stream_version, which will be used to check with the `expected_version` when appending events, is **5**, causing a `wrong_expected_version` error when the aggregate tries to persist newly created events.

## Is this a bug or misuse
If this behavior is not expected, I think it's Commanded that takes wrong stream_version. Or simply I should never use such an approach to migrate. Any ideas?

[...]

PR #355, opened 2020-03-20
Fix calculation of aggregate_version

CptBreeza:
Address https://github.com/commanded/commanded/issues/353


slashdotdash:
Thanks for the pull request. I want to consider if there are any other implications of using `event_number` instead of `stream_version` as the aggregate's version before merging.

CptBreeza:
> Thanks for the pull request. I want to consider if there are any other implications of using `event_number` instead of `stream_version` as the aggregate's version before merging.

Thanks.

vasspilka:
We will not be using same streams across different eventstores ever correct?
For new applications this would not be a problem at all, would running applications have a chance to process events not in order if they where to upgrade?

CptBreeza:
So is there any progress on this?

PR #356, opened 2020-03-26
Transient event handler proof-of-concept

evanreierson:
#### Summary
This PR adds a `:projection` option to the `Commanded.Event.Handler` module. The `:projection` option can be either `:persistent` (default) or `:in_memory`.

The `:persistent` option does not change the behaviour of the current `Commanded.Event.Handler`, but when `:in_memory` is applied, event handlers receive all their events each time they start. This allows developers to create in-memory projections using OTP primitives.

I also implemented a separate `Commanded.Event.TransientHandler` module (not shown in PR), but since the behaviours almost entirely overlap, I figured this approach would have a much higher maintainability cost.

#### Outstanding questions
How can the tests/test coverage for this feature be improved?
How is the naming? Is `@type projection :: :persistent | :in_memory` intuitive?
I'm no expert in Commanded or Elixir, does it look like I introduced any concurrency bugs?

#### Potential shortcomings
Having multiple handlers from the same module subscribed to the same stream is not supported.

#### To do
I still need to update the documentation and improve tests/test coverage.
\
\
Looking forward to any feedback!

### DIFF
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -243,6 +243,7 @@ defmodule Commanded.Event.Handler do
   @type metadata :: map()
   @type subscribe_from :: :origin | :current | non_neg_integer()
   @type consistency :: :eventual | :strong
+  @type projection :: :persistent | :in_memory
 
   @doc """
   Optional initialisation callback function called when the handler starts.
@@ -387,7 +388,8 @@ defmodule Commanded.Event.Handler do
       module_opts
       |> Keyword.merge(local_opts)
       |> Keyword.split(
-        [:application, :consistency, :start_from, :subscribe_to] ++ additional_allowed_opts
+        [:application, :consistency, :start_from, :subscribe_to, :projection] ++
+          additional_allowed_opts
       )
 
     if Enum.any?(invalid) do
@@ -421,11 +423,16 @@ defmodule Commanded.Event.Handler do
     :subscribe_from,
     :subscribe_to,
     :subscription,
-    :subscription_ref
+    :subscription_ref,
+    :projection
   ]
 
   @doc false
   def start_link(application, handler_name, handler_module, opts \\ []) do
+    projection = projection(opts)
+    subscribe_to = subscribe_to(opts)
+
+    handler_name = handler_name(handler_name, subscribe_to, projection)
     name = name(application, handler_name)
 
     handler = %Handler{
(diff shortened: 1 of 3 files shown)


lib/commanded/event/handler.ex
```diff
@@ -761,15 +792,22 @@ defmodule Commanded.Event.Handler do
     case opts[:start_from] || :origin do
       start_from when start_from in [:origin, :current] -> start_from
       start_from when is_integer(start_from) -> start_from
-      invalid -> "Invalid `start_from` option: #{inspect(invalid)}"
```
evanreierson: Is there a reason this did not have a raise or is it a bug?

lib/commanded/event/handler.ex
```diff
@@ -761,15 +792,22 @@ defmodule Commanded.Event.Handler do
...
+      invalid -> raise "Invalid `start_from` option: #{inspect(invalid)}"
     end
   end
 
   defp subscribe_to(opts) do
     case opts[:subscribe_to] || :all do
       :all -> :all
       stream when is_binary(stream) -> stream
-      invalid -> "Invalid `subscribe_to` option: #{inspect(invalid)}"
```
evanreierson: No raise here either

lib/commanded/event/handler.ex
slashdotdash: This is a bug, good catch.

lib/commanded/event/handler.ex
slashdotdash: Ditto, it's a bug.

slashdotdash:
Thanks for the pull request Evan. 

Looking at the changes you've made I think we should use a different name for the `projection` option as it's really about whether we want the event handler to start a brand new subscription each time. 

Currently an event handler will always resume a subscription if it exists. This change would allow a new subscription to be created each time it starts. Therefore the choice should be either `:resume` (default) or `:restart`. I haven't yet come up with a name for the option, I'm open to suggestions.

Ideally we also want the subscription to be deleted on `terminate`.

evanreierson:
Hmm how about `@type subscription :: :resume | :restart`? There is a `:subscription` field in the handler struct though so that could be confusing... How do you feel about a boolean option? `@type restart_subscription :: true | false`?

Also, any preference between subscription option `:restart` and `:reset`?  I think both get the point across.

What about the lack of guarantees on `terminate`? If the server gets struck by lightning, or more likely, the developer hits `ctrl + c` in IEx, `terminate` will not run. That's why I chose to delete the subscription (when the `:restart` option is chosen) before a new one is created.

tcoopman:
Sorry to comment on a closed PR, but I was wondering why it was closed without further comments?
I was just looking for a feature like this where you could have "transient" event handlers that receive the events from the start every time they restart.

Is there an other way to achieve this?

PR #359, opened 2020-03-27
Test and document dispatch's :function option

jwilger:
There is an existing yet undocumented option to the
`Commanded.Commands.Router`'s `dispatch` macro that allows you to
specify the name of the function that will be called on the aggregate to
which the command is being dispatched (i.e. you can use a method other
than `execute/2` on your aggregate).

Using this option allows one to have more specifically-named functions on
an aggregate (vs. just pattern matching `execute/2` based on the type of
command, etc.) without requiring the use of a command handler module
when one is otherwise unnecessary. This results in fewer lines of code
as well as fewer layers of abstraction while allowing for the option of
self-documenting function names.

Although the necessary behavior was already in place, because it was
neither documented nor directly tested, it wouldn't be considered part
of the official API, and it could be dangerous to rely on that
capability (depsite the fact that such seems to have been stable for the
past three years or so). This change simply adds test coverage and
documentation for this option's use.

### DIFF
--- a/guides/Commands.md
+++ b/guides/Commands.md
@@ -87,11 +87,23 @@ It is also possible to route a command directly to an aggregate, without requiri
 defmodule BankRouter do
   use Commanded.Commands.Router
 
+  # will route to `BankAccount.execute/2`
   dispatch OpenAccount, to: BankAccount, identity: :account_number
 end
 ```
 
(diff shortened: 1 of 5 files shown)


slashdotdash:
Thanks for the pull request John. Specifying the aggregate function other than `execute/2` wasn't an intended usage, but it makes sense to officially support it.

PR #360, opened 2020-03-29
Clarify execution context for command handlers

jwilger:
Updates documentation to clarify that the execution context of command
handlers is the same as the aggregate; the command handler is executed
in the same GenServer process and is within the consistency boundary of
the aggregate.


slashdotdash:
Thanks John.

PR #363, opened 2020-04-14
Remove router module compile-time checking

### LINKED ISSUE #358: Misleading error message when file has a syntax error
I have run into this twice where the overloaded `Code.ensure_compiled?` error is overloaded in Commanded and displays a canned message about missing aliases.

For example, I was missing an alias in this case for `AuditDenied` and `DenyAudit`, but the error makes you think something is wrong in the router, not in the aggregate.

```elixir
defmodule Capacitor.Aggregates.Audit do
  alias Capacitor.Aggregates.Audit
  alias Capacitor.Commands.{ApproveAudit, CreateAudit}
  alias Capacitor.Events.{AuditApproved, AuditCreated}

 # ...
  def execute(%Audit{status: :open}, %DenyAudit{} = command) do
    AuditDenied.new(command)
  end
end
```

Here is the error I received trying to run `mix compile`

Screen Shot 2020-03-27 at 9 37 23 AM

This sends you to the `router.ex` to look for a missing alias only to find it's another file.

---

I had a similar issue yesterday but that time it was a typo in one of my Commands. The only way I was able to find it was to comment out chunks of code and find what made `mix compile` pass.

Is there any way to make this error more friendly to point to where the issue in fact is?

slashdotdash:
Compile-time checking for router dependencies causes compilation errors to be hidden, making it difficult to locate the error to be fixed. As an example a compiler error with an aggregate module simply returns the error message: "module `Aggregate` does not exist".  

This pull request removes the compile-time checking and reworks the `Commanded.Commands.Router.dispatch` macro to work without attempting to load the dependent modules. 

Fixed #358.


jsmestad:
This works like a charm. Thank you!

PR #368, opened 2020-04-30
Fix high memory consumption during aggregate rebuild

EskiMag:
I tried benchmarking both solutions and the time it takes is around 60% of the old implementation.
Even bigger difference is in the amount of memory being allocated and not released during the rebuild. With old implementation it gradualy rises into gigabytes. With new implementation it stays around 100-200 megabytes. I am not sure if the Stream is not used correctly, or if there's some bug in the Stream internals or if it's just my special case, but it could be useful if somebody could do some benchmarks with his data to prove my suspicion.

**Before change:**
Screenshot 2020-04-30 at 10 52 08
Screenshot 2020-04-30 at 10 52 05

**After change:**
Screenshot 2020-04-30 at 12 03 30
Screenshot 2020-04-30 at 12 03 24

I tested this on around `100_000` events stored as JSONB in PostgreSQL.

Benchmarking machine:
```
Operating System: macOS
CPU Information: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
Number of Available Cores: 4
Available memory: 16 GB
Elixir 1.10.2
Erlang 22.2.8
```

I didn't write new test(s). I assume, this is covered by `event_persistence_test.exs`.

### DIFF
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -399,7 +399,7 @@ defmodule Commanded.Aggregates.Aggregate do
     |> Stream.map(fn event ->
       {event.data, event.stream_version}
     end)
-    |> Stream.transform(state, fn {event, stream_version}, state ->
+    |> Enum.reduce_while(state, fn {event, stream_version}, state ->
       case event do
         nil ->
           {:halt, state}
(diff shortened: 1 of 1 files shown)


slashdotdash:
This looks great, thank you @EskiMag.

PR #380, opened 2020-05-25
Capture exception on Process Manager apply/2 and call error/3 callback

### LINKED ISSUE #379: Catch exceptions in Process Manager apply/2
Hi,

I have a PM that has defined the `error/3` callback but when throwing an exception during the `apply/2` command does not trigger the `errors/3` callback

Looking at the code I figured it out that we are not rescuing on apply/2

Handle processing:
https://github.com/commanded/commanded/blob/4a6df4a3fc3e318805574ca9cab5cef0a83617e3/lib/commanded/process_managers/process_manager_instance.ex#L235-L243

Apply processing:
https://github.com/commanded/commanded/blob/4a6df4a3fc3e318805574ca9cab5cef0a83617e3/lib/commanded/process_managers/process_manager_instance.ex#L323

The `interested/3` is also catching exceptions:
https://github.com/commanded/commanded/blob/4a6df4a3fc3e318805574ca9cab5cef0a83617e3/lib/commanded/process_managers/process_router.ex#L292-L398

If you think this is something that we should fix I can work on a PR.

marpo60:
Closes #379

### DIFF
--- a/lib/commanded/process_managers/process_manager_instance.ex
+++ b/lib/commanded/process_managers/process_manager_instance.ex
@@ -203,18 +203,24 @@ defmodule Commanded.ProcessManagers.ProcessManagerInstance do
         opts = [causation_id: event_id, correlation_id: correlation_id, returning: false]
 
         with :ok <- commands |> List.wrap() |> dispatch_commands(opts, state, event) do
-          process_state = mutate_state(event, state)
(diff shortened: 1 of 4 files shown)


slashdotdash:
Thanks @marpo60.

PR #383, opened 2020-06-05
Fix aggregate multis not returning intermediate events

y86:
Events returned by `Multi.execute` calls are not persisted if followed by `Multi.execute` calls that return another `multi` instead of events.


slashdotdash:
@y86 Could you extract the process manager cleanup changes into a separate pull request, leaving this one only contain the Multi fix.

y86:
I submitted a new one with only the multi changes

PR #385, opened 2020-06-17
Fix multi aggregate

y86:
Events returned by `Multi.execute` calls are not persisted if followed by `Multi.execute` calls that return another `multi` instead of events.


slashdotdash:
Thanks for fixing this bug @y86.

Issue #339, opened 2020-02-11
Pass enriched metadata to upcaster protocol


slashdotdash:
The `Commanded.Event.Upcaster` protocol allows events to be upcast by providing the source event and its metadata.  It would be useful to have access to the additional fields provided by the `Commanded.EventStore.RecordedEvent` struct, such as the event's `created_at` timestamp. 

Currently event metadata enrichment is only [applied to event handlers](https://github.com/commanded/commanded/blob/v1.0.0/lib/commanded/event/handler.ex#L681). Instead, this enrichment could be moved before upcasting to make it available to any event processors.

### Example usage

```elixir
defimpl Commanded.Event.Upcaster, for: AnEvent do
  def upcast(%AnEvent{} = event, metadata) do
    %AnEvent{name: name} = event
    %{:created_at => created_at} = metadata
    
    IO.inspect(metadata)
    # %{
    #   :causation_id => "db1ebd30-7d3c-40f7-87cd-12cd9966df32",
    #   :correlation_id => "1599630b-9c38-433c-9548-0dd793108ba0",
    #   :created_at => #DateTime<2017-10-30 11:19:56.178901Z>,
    #   :event_id => "5e4a0f38-385b-4d57-823b-a1bcf705b7bb",
    #   :event_number => 12345,
    #   :stream_id => "e42a588d-2cda-4314-a471-5d008cce01fc",
    #   :stream_version => 1,
    #   "issuer_id" => "0768d69a-d2b7-48f4-d0e9-083a97f7ebe0",
    #   "user_id" => "user@example.com"
    # }

    %AnEvent{event | first_name: name, created_at: created_at}
  end
end
```

vasspilka:
Hey, can I open a PR for this?

slashdotdash:
@vasspilka Yes, please do.

vasspilka:
@slashdotdash PR oppened

slashdotdash:
Fixed by #389.

Issue #327, opened 2019-12-10
Process Manager Metadata for Commands 


arfl:
I am curious: is it some way for process managers to add metadata to the dispatched commands. The `handle` function is supposed to return one command or a list of commands to be dispatched...

slashdotdash:
Unfortunately it's not currently supported. 

Can you give me an example of what kind of metadata you want to include? We could extend to process manager API to allow setting metadata.

trbngr:
One example: I'd love to be able to set the correlation id

arfl:
We do not use the causation_id/correlation_id provided by commanded/eventstore because they are just uuids. Our IDs are URIs and look like this one: `tech.studitemps:my_bounded_context_name.my_event_name.e7f4e458-3da6-427d-88e3-227bbad22062`, where `studitemps.tech` is our web site. So just looking at this kind of IDs you can determine what kind of event it is and which bounded context has thrown this event.
So we would like to store those causation_ids/correlation_ids in process managers also.

slashdotdash:
Pull request #389 provides access to enriched metadata in the [`Commanded.Event.Upcaster` protocol](https://hexdocs.pm/commanded/Commanded.Event.Upcaster.html). This could be used to copy any fields from the [`Commanded.EventStore.RecordedEvent`](https://hexdocs.pm/commanded/Commanded.EventStore.RecordedEvent.html) struct, such as the causation and correlation ids.

jdav-dev:
My team has a use case for this.  We store a `user_id` in metadata.  We would like process managers to take this `user_id` from events it handles and add it to commands it dispatches.  Right now, process managers cannot add metadata, nor do they receive metadata with the events.  Our workaround is middleware that copies the `user_id` based on a command's `causation_id`.

edwardzhou:
unable to access metadata not only in process manager, but also within Aggregate. Since Router.dispatch receives metadata in 3rd parameter, it is reasonable to pass it Aggregate.execute. this is very useful to carry the request user_id through the execution flow.

nicklayb:
Similarly to @jdav-dev,  we need metadata to be forwarded to next commands. We are using a commanded middleware to authorize command originators to dispatch commands. Pretty basic middleware that checks if the command's metadata includes an originator and that the originator is allowed to perform such command. The thing is that the initial command's metadata aren't forwarded to subsequent therefore following commands aren't authorized.

[...]

PR #389, opened 2020-07-05
Allow metadata in upcaster

vasspilka:
We allowing the Upcaster protocol casses to RecordedEvent metadata regarding the event.
This will allow to solve idepotency issues on EventHandlers and ProcessManagers by allowing to do checks based `event_uuid` or `event_number` as well as having access to `created_at` fields and correlation/causation, stream info.

This PR addresses #339 directly but also indirectly solves #327

### DIFF
--- a/test/event/upcaster_test.exs
+++ b/test/event/upcaster_test.exs
@@ -6,7 +6,7 @@ defmodule Event.UpcasterTest do
   alias Commanded.EventStore
   alias Commanded.EventStore.EventData
   alias Commanded.EventStore.RecordedEvent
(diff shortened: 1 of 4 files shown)


test/event/upcaster_test.exs
```diff
@@ -30,6 +30,27 @@ defmodule Event.UpcasterTest do
       assert version == 2
     end
 
+    test "upcasted event includes event metadata" do
```
vasspilka, line 33:
```suggestion
    test "upcasted event can include event metadata" do
```

slashdotdash:
Great stuff, thanks @vasspilka

PR #391, opened 2020-07-27
Add support to event store subscription options

YgorCastor:
There are some events store backends that allows some persistent subscriptions tuning, right now there's no way to configure those since the base behavior only cares about start_from. This pull request adds a new optional callback that allows the user to send a Keyword list with custom adapter-specific options.


slashdotdash:
Thanks for the pull request @YgorCastor. I will get this merged in before the next Commanded release.

slashdotdash:
Sorry, this has been automatically closed due to me changing the default branch from `master` to `develop`. An unintended side-effect!

slashdotdash:
Merged in 0c53a05993d5ad57ba0dd88ffd9a6a19c978b55e.

PR #400, opened 2020-08-16
Support custom state for event handlers

### LINKED ISSUE #366: Support Custom State for Event Handlers?
Background: We are building a platform to support multiple e-commerce stores. Each store needs to sync some of its data to an external source/API. Each store should run independently from each other, a failure in some store should not impact any other store. Therefore each store (variable number on our platform) should have one event handler to sync data. 

Each selected event in the stream needs to successfully be sync'd with the external API before moving on to the next event in order to guarantee successful processing & correct order. 

Issue: `Commanded.Event.Handler` does not support custom state to be passed upon start up all the way down to `handle` to allow filter & process of the correct events in order to accomplish the sync per store in our case. Nor a dynamic `name` to support multiple processes of the same event handler module. 

Some rough code idea: 

```elixir
defmodule Sync do
  use Commanded.Event.Handler, application: App, ...
  def handle(%AnEvent{..}, %{state: %{store: "foo"}} = _metadata), do: ...
end

Sync.start_link(name: "Sync#foo", store: "foo")
```

Would something like this make sense to be supported? Can I accomplish my objective differently?

Thank you!

slashdotdash:
An event handler can define and update custom state which is held in the `GenServer` process' memory. It is passed to the `handle/2` function as part of the metadata using the `:state` key. The state is transient and will be lost whenever the process restarts. 

Initial state can be set in the `init/1` callback function by adding a `:state` key to the config. It can also be provided when starting the handler process:

```elixir
ExampleHandler.start_link(state: initial_state)
```

Or when supervised:

```elixir
Supervisor.start_link([
  {ExampleHandler, state: initial_state}
], strategy: :one_for_one)
```

State can be updated by returning `{:ok, new_state}` from any `handle/2` function. Returning an `:ok` reply will keep the state unchanged.

Handler state is also included in the `Commanded.Event.FailureContext` struct passed to the `error/3` callback function.

### Example

```elixir
defmodule StatefulEventHandler do
  use Commanded.Event.Handler,
    application: ExampleApp,
    name: __MODULE__

  def init(config) do
    config = Keyword.put_new(config, :state, %{})

    {:ok, config}
  end

  def handle(event, metadata) do
    %{state: state} = metadata

    new_state = mutate_state(state)

    {:ok, new_state}
  end
end
```

Closes #366.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Allow `GenServer` start options to be provided when starting event handlers and process managers ([#398](https://github.com/commanded/commanded/pull/398)).
 - Add `hibernate_after` option to application config ([#399](https://github.com/commanded/commanded/pull/399)).
 - Add support for providing adapter-specific event store subscription options ([#391](https://github.com/commanded/commanded/pull/391)).
+- Support custom state for event handlers ([#400](https://github.com/commanded/commanded/pull/400)).
 
 ## v1.1.1
 
--- a/lib/commanded/event/failure_context.ex
+++ b/lib/commanded/event/failure_context.ex
@@ -4,19 +4,27 @@ defmodule Commanded.Event.FailureContext do
 
   The available fields are:
 
-    - `application` - the associated `Commanded.Application`.
-    - `handler_name` - the name of the event handler.
-    - `context` - a map that is passed between each failure. Use it to store any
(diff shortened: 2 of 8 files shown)


larskluge:
Awesome, thank you @slashdotdash !

PR #402, opened 2020-08-24
Pass execution context to aggregates before execution

cjuniet:
Add an optional `before_execute/2` callback to Aggregates, called just before the `execute/2` callback, with the user-defined aggregate state and the current execution context. This is particularly useful to do some setup in the context of the aggregate process (e.g. set the aggregate's own logger's metadata).

See issue https://github.com/commanded/commanded/issues/403


slashdotdash:
Thanks for the PR @cjuniet.

Issue #387, opened 2020-06-27
Telemetry/Tracing instrumentation


davydog187:
We have begun instrumenting our application with telemetry and started pushing traces to DataDog APM. We are achieving this currently with https://github.com/spandex-project/spandex. Currently, we have a commanded middleware that generates a trace for each command, and then an event handler that creates spans for each event emitted. Traces end up looking like the following.

<img width="1131" alt="spans" src="https://user-images.githubusercontent.com/1019721/85925475-a1443c00-b866-11ea-9d8d-4142bdc4e228.png">

Notice that the event times our skewed way past the end of the command? Since the event handler is the one creating the child spans, it is victim to head of line blocking that makes the span inaccurate. Below is some sample code for how the event spans are created

```elixir
defmodule MyApp.EventHandlers.Spandex do
  use Commanded.Event.Handler,
    ...

  def handle(%type{}, metadata) do
    %{"trace_id" => trace_id, "parent_id" => parent_id, created_at: created_at} = metadata
    # start the span at the created at time of the event
    start = DateTime.to_unix(created_at, :nanosecond)

    # Continue this trace from the one set up in the middleware
    context = %Spandex.SpanContext{trace_id: trace_id, parent_id: parent_id}
    Tracer.continue_trace(Util.module_name(type, 4), context)

    # Set the start time
    Tracer.update_span(start: start)

    Complete the trace
    Tracer.finish_trace()
    :ok
  end
end
```

Since we don't have information about the end time of the trace, what happens is that we get inaccurate end times. This is better than nothing, but I think there is an opportunity to improve it. I wanted to create this issue to see if anyone else has thought more about this problem and if there is a way to potentially hook deeper into commanded so we can get more accurate spans.

An additional problem that we're dealing with is that its hard to instrument something like a process manager for tracing. As you can see in my example, I use the pipeline metadata to propagate trace and span ids. Without putting these values directly into our commands and domain events, there isn't a great way to propagate this information to a process manager or aggregate to include spans for it in the trace. If commanded instrumented itself with telemetry, this could be possible by simply adding a telemetry handler.

davydog187:
This is what our middleware looks like

```elixri
defmodule MyApp.Spandex.Commanded.Middleware do
  ...

  def before_dispatch(%Pipeline{} = pipeline) do
    %Pipeline{command: command} = pipeline

    Tracer.start_trace(Util.module_name(command, 4))

    pipeline
[...]

PR #407, opened 2020-10-29
 start/stop Aggregate telemetry events

davydog187:
Screen Shot 2020-10-29 at 2 44 18 PM

Contributes to #387

Adds two telemetry events to the Aggregate GenServer

1. [:commanded, :aggregate, :execute, :start]
2. [:commanded, :aggregate, :execute, :stop]

Each provide the `%ExecutionContext{}` and `%Aggregate{}` as metadata.
This allows metrics and tracing instrumentation elsewhere.

As an example of how this can be useful, I have built a small working
prototype here.

https://github.com/davydog187/foosball

### DIFF
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -1,4 +1,22 @@
 defmodule Commanded.Aggregates.Aggregate do
+  use TelemetryRegistry
+  use GenServer, restart: :temporary
+  use Commanded.Registration
+
+  telemetry_event(%{
+    event: [:commanded, :aggregate, :execute, :start],
+    description: "Emitted when an aggregate starts executing a command",
+    measurements: "%{system_time: integer()}",
+    metadata: "%{execution_context: ExecutionContext.t(), aggregate: %__MODULE__{}}"
+  })
+
+  telemetry_event(%{
+    event: [:commanded, :aggregate, :execute, :stop],
+    description: "Emitted when an aggregate stops executing a command",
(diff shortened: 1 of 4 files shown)


davydog187:
Screen Shot 2020-10-29 at 2 04 54 PM

lib/commanded/aggregates/aggregate.ex
```diff
@@ -242,22 +267,39 @@ defmodule Commanded.Aggregates.Aggregate do
...
+          {:stop, reason} -> {:stop, reason, reply, state}
+          lifespan_timeout -> {:reply, reply, state, lifespan_timeout}
+        end
       end
-    end
+
+    duration = System.monotonic_time() - start
+
+    :telemetry.execute(
+      [:commanded, :aggregate, :execute, :stop],
```
davydog187, line 297: my only reservation here is if there should be a separate event for errors

davydog187:
@slashdotdash I should also mention that we are now using Commanded in production at https://simplebet.io/, so you can update the wiki! It's powering a backend service that supports https://www.fanduel.com/fanduelplayaction

lib/commanded/aggregates/aggregate.ex
slashdotdash: It would be useful to capture exceptions/errors, but that can be added in a separate PR.

slashdotdash:
Great stuff, thanks Dave.

lib/commanded/aggregates/aggregate.ex
davydog187: noted.

PR #408, opened 2020-10-29
start/stop/error Event.Handler telemetry events

davydog187:
Follows #407 (and currently branched off of)

Contributes to #387 

Adds telemetry to `Event.Handler.handle/2`. For this one, I made a separate event for error, which aligns with https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/router.ex#L345

I will backport the error event to the other PR

### DIFF
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -3,6 +3,45 @@ defmodule Commanded.Event.Handler do
   use Commanded.Registration
   use TelemetryRegistry
 
+  telemetry_event(%{
+    event: [:commanded, :event, :handle, :start],
+    description: "Emitted when an Event.Handler.handle/2 callback is started",
+    measurements: "%{system_time: integer()}",
+    metadata: """
+    %{
+      recorded_event: RecordedEvent.t(),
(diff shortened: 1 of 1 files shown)


lib/commanded/event/handler.ex
```diff
@@ -748,31 +792,71 @@ defmodule Commanded.Event.Handler do
...
+
+    :telemetry.execute([:commanded, :event, :handle, :stop], measurements, metadata)
+
+    case handled_event do
       :ok ->
+        :telemetry.execute([:commanded, :event, :handle, :stop], measurements, metadata)
         confirm_receipt(event, state)
 
       {:ok, handler_state} ->
+        :telemetry.execute([:commanded, :event, :handle, :stop], measurements, metadata)
```
slashdotdash, line 815: The updated `handler_state` should be included in the telemetry `metadata`.

slashdotdash (commented):
One minor comment, otherwise looks good.

lib/commanded/event/handler.ex
davydog187: @slashdotdash updated.

PR #417, opened 2020-12-04
Update elixir version for github actions

davydog187:
My other PRs are failing on the elixir install. Hopefully updating will help

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,6 @@ on: [push, pull_request]
 
 jobs:
   build:
-
     name: Build and test
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -4,13 +4,12 @@ on: [push, pull_request]
...
 jobs:
   build:
-
     name: Build and test
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        otp: ['22.3']
-        elixir: ['1.10.4']
+        otp: ['23.1']
```
slashdotdash, line 11: We still want to test with Erlang 22.3.x as well:

```suggestion
        otp: ['22.3', '23.1']
```

davydog187: Done

PR #418, opened 2020-12-07
Telemetry [:commanded, :process_manager, :handle]

davydog187:
Adds telemetry events for Process Managers.

Contributes to https://github.com/commanded/commanded/issues/387

### DIFF
--- a/lib/commanded/process_managers/process_manager.ex
+++ b/lib/commanded/process_managers/process_manager.ex
@@ -1,4 +1,41 @@
 defmodule Commanded.ProcessManagers.ProcessManager do
+  use TelemetryRegistry
+
+  telemetry_event(%{
+    event: [:commanded, :process_manager, :handle, :start],
+    description: "Emitted when a process manager starts handling an event",
+    measurements: "%{system_time: integer()}",
+    metadata: """
+    %{application: Commanded.Application.t(),
+      process_manager_name: String.t() | Inspect.t(),
+      process_manager_module: module(),
+      process_state: term(),
+      process_uuid: String.t()}
+    """
+  })
+
+  telemetry_event(%{
+    event: [:commanded, :process_manager, :handle, :stop],
+    description: "Emitted when a process manager stops handling an event",
+    measurements: "%{system_time: integer()}",
+    metadata: """
+    %{application: Commanded.Application.t(),
+      process_manager_name: String.t() | Inspect.t(),
+      process_manager_module: module(),
+      process_state: term(),
+      process_uuid: String.t()}
+    """
+  })
+
+  telemetry_event(%{
+    event: [:commanded, :process_manager, :handle, :exception],
+    description: "Emitted when a process manager raises an exception",
+    measurements: "%{system_time: integer()}",
+    metadata: """
+    %{application: Commanded.Application.t()}
+    """
+  })
+
   @moduledoc """
   Macro used to define a process manager.
 
@@ -219,6 +256,11 @@ defmodule Commanded.ProcessManagers.ProcessManager do
 
   The above example requires three named Commanded applications to have already
   been started.
+
+  ## Telemetry
+
+  #{telemetry_docs()}
+
   """
 
   alias Commanded.ProcessManagers.FailureContext
--- a/lib/commanded/process_managers/process_manager.ex
+++ b/lib/commanded/process_managers/process_manager.ex
@@ -20,6 +20,8 @@ defmodule Commanded.ProcessManagers.ProcessManager do
     measurements: "%{system_time: integer()}",
     metadata: """
     %{application: Commanded.Application.t(),
+      commands: [struct()],
+      error: nil | any(),
       process_manager_name: String.t() | Inspect.t(),
       process_manager_module: module(),
       process_state: term(),
--- a/lib/commanded/process_managers/process_manager.ex
+++ b/lib/commanded/process_managers/process_manager.ex
@@ -32,7 +32,14 @@ defmodule Commanded.ProcessManagers.ProcessManager do
     description: "Emitted when a process manager raises an exception",
     measurements: "%{system_time: integer()}",
     metadata: """
-    %{application: Commanded.Application.t()}
+    %{application: Commanded.Application.t(),
+      process_manager_name: String.t() | Inspect.t(),
+      process_manager_module: module(),
+      process_state: term(),
+      process_uuid: String.t(),
+      kind: :throw | :error | :exit,
+      reason: any(),
+      stacktrace: list()}
     """
   })
 
--- a/lib/commanded/process_managers/process_manager_instance.ex
+++ b/lib/commanded/process_managers/process_manager_instance.ex
@@ -180,9 +180,8 @@ defmodule Commanded.ProcessManagers.ProcessManagerInstance do
 
     :ok = ack_event(event, state)
 
-    {:noreply, state, idle_timeout}
-
     telemetry_stop(start_time, telemetry_metadata, [])
+
     {:noreply, state, idle_timeout}
   end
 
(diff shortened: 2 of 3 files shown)


test/process_managers/process_manager_telemetry_test.exs
```diff
@@ -0,0 +1,232 @@
...
+  # TODO change to a local example
+
+  describe "process manager telemetry" do
+    test "emit `[:commanded, :process_manager, :handle, :start]` event" do
+      transfer_uuid = UUID.uuid4()
+      debit_account = UUID.uuid4()
+      credit_account = UUID.uuid4()
+      expected_source_uuid = "\"TransferMoneyProcessManager\"-\"#{transfer_uuid}\""
+
+      expect(MockEventStore, :read_snapshot, fn _adapter_meta, ^expected_source_uuid ->
```
davydog187, line 40: I'm going to rework these to use a local example.

lib/commanded/process_managers/process_manager.ex
```diff
@@ -1,4 +1,48 @@
...
+  telemetry_event(%{
+    event: [:commanded, :process_manager, :handle, :stop],
+    description: "Emitted when a process manager stops handling an event",
+    measurements: "%{system_time: integer()}",
+    metadata: """
+    %{application: Commanded.Application.t(),
+      process_manager_name: String.t() | Inspect.t(),
+      process_manager_module: module(),
+      process_state: term(),
+      process_uuid: String.t()}
```
davydog187, line 26: this also passes the commands dispatched

davydog187:
Some feedback from @slashdotdash a few weeks ago

```
Ben Smith @slashdotdash Dec 09 15:23
Did you look at what happens when events are retried on error as the handle_event_error function calls back into the process_unseen_event function when the error callback returns :retry?

Looking at how the telemetry is implemented I think it would do:

[:commanded, :process_manager, :handle, :start] (1st attempt)
[:commanded, :process_manager, :handle, :start] (2nd attempt)
[:commanded, :process_manager, :handle, :start] (3rd attempt)
[:commanded, :process_manager, :handle, :start] (Nth attempt)

[:commanded, :process_manager, :handle, :stop | :exception] (Nth attempt)

[:commanded, :process_manager, :handle, :stop | :exception] (3rd attempt)
[:commanded, :process_manager, :handle, :stop | :exception] (2nd attempt)
[:commanded, :process_manager, :handle, :stop | :exception] (1st attempt)

Dave Lucia @davydog187 Dec 09 15:23
i didn't, its a good point
i'll add some tests to validate, what would you expect to happen if that occurs?

Ben Smith @slashdotdash Dec 09 15:27
The event handler telemetry does the following:
[:commanded, :process_manager, :handle, :start] (1st attempt)
[:commanded, :process_manager, :handle, :stop | :exception] (1st attempt)
[:commanded, :process_manager, :handle, :start] (2nd attempt)
[:commanded, :process_manager, :handle, :stop | :exception] (2nd attempt)
[:commanded, :process_manager, :handle, :start] (Nth attempt)
[:commanded, :process_manager, :handle, :stop | :exception] (Nth attempt)

Dave Lucia @davydog187 Dec 09 15:27


Ben Smith @slashdotdash Dec 09 15:27
The telemetry span covers each attempt and stop (error) or exception
It might be worth including command dispatch in the Process manager telemetry too

Dave Lucia @davydog187 Dec 09 15:28
it does on the stop event
i think i missed that in the docs

Ben Smith @slashdotdash Dec 09 15:28
Both event handling and invidual command dispatches can be failed and retried

Dave Lucia @davydog187 Dec 09 15:28
i'll add a comment
oh i see what you mean
having events just for the dispatch
i was thinking that the dispatcher should handle that, no?

Ben Smith @slashdotdash Dec 09 15:30
It could be:

[:commanded, :process_manager, :handle, :start]
for each command:

[:commanded, :process_manager, :dispatch, :start]
[:commanded, :process_manager, :dispatch, :stop]
after dispatching

[:commanded, :process_manager, :handle, :stop]
Since handling an event for a process manager instance covers the event handling and all command dispatching
```

lib/commanded/process_managers/process_manager_instance.ex
```diff
@@ -173,9 +175,15 @@ defmodule Commanded.ProcessManagers.ProcessManagerInstance do
   defp process_seen_event(%RecordedEvent{} = event, %State{} = state) do
     %State{idle_timeout: idle_timeout} = state
 
+    telemetry_metadata = telemetry_metadata(event, state)
+    start_time = telemetry_start(telemetry_metadata)
+
     :ok = ack_event(event, state)
 
     {:noreply, state, idle_timeout}
```
davydog187, line 183: remove

lib/commanded/process_managers/process_manager_instance.ex
```diff
@@ -184,45 +191,56 @@ defmodule Commanded.ProcessManagers.ProcessManagerInstance do
...
+
+        handle_event_error({:error, error}, event, failure_context, state)
 
       {:error, error, stacktrace} ->
         failure_context = build_failure_context(event, context, stacktrace, state)
 
-        handle_event_error({:error, error}, event, failure_context, state)
+        telemetry_stop(start_time, telemetry_metadata, {:error, error, stacktrace})
 
-      {:stop, _error, _state} = reply ->
```
davydog187, line 198: I'm not sure why this was here previously, it's not part of the `handle/2` API

lib/commanded/process_managers/process_manager_instance.ex
```diff
@@ -184,45 +191,56 @@ defmodule Commanded.ProcessManagers.ProcessManagerInstance do
...
-        with :ok <- commands |> List.wrap() |> dispatch_commands(opts, state, event) do
-          case mutate_state(event, state) do
-            {:error, error, stacktrace} ->
-              failure_context = build_failure_context(event, context, stacktrace, state)
+        commands
+        |> List.wrap()
+        |> dispatch_commands(opts, state, event)
+        |> case do
+          :ok ->
+            telemetry_stop(start_time, telemetry_metadata, commands)
```
davydog187, line 221: I edited this slightly as the previous formatting was tripping my brain for some reason. I can change back if you prefer the `with`

test/process_managers/process_manager_telemetry_test.exs
```diff
@@ -0,0 +1,316 @@
...
+                 error: "stop"
+               },
+               metadata
+             )
+
+      refute_receive {[:commanded, :process_manager, :handle, :exception], _num, _measurements,
+                      _metadata}
+    end
+
+    test "events are emitted with discrete start/stop on retries" do
```
davydog187, line 199: this test doesn't cover all retry cases, covering every one would be a bit of a burden, but i'm happy to do it if we think its worthwhile.

test/process_managers/process_manager_telemetry_test.exs
```diff
@@ -0,0 +1,316 @@
...
+
+    :telemetry.attach_many(
+      @handler,
+      [
+        [:commanded, :process_manager, :handle, :start],
+        [:commanded, :process_manager, :handle, :stop],
+        [:commanded, :process_manager, :handle, :exception]
+      ],
+      fn event_name, measurements, metadata, reply_to ->
+        num = Agent.get_and_update(agent, fn num -> {num, num + 1} end)
```
davydog187, line 288: I decorated with an incrementing number here so i could make sure the `assert_receive`'s were being received in the right order. Without this you don't get that guarantee, so the tests were misleading

lib/commanded/process_managers/process_manager_instance.ex
slashdotdash: This would match the `{:stop, reason :: term(), new_state}` return value for [`handle_cast/2`](https://hexdocs.pm/elixir/GenServer.html#c:handle_cast/2) where the error is given as the reason to stop the process.

PR #419, opened 2020-12-14
Update Usage.md

jannikbecher:
Add application option to Commanded.Commands.Router


slashdotdash:
Router modules _do not_ need to reference the Commanded application as commands are meant to be dispatched via the application, not the router.

PR #420, opened 2020-12-23
WIP detect failed execution due to nodedown

drteeth:
Retry failed command dispatch due to the registered node for an aggregate being down when we try to execute against it

- [x] identify the execution path 
- [ ] implement the retry


drteeth:
Oops. not quite ready. Sorry.

PR #423, opened 2020-12-29
Telemetry [:commanded, :application, :dispatch]

davydog187:
Adds telemetry to the dispatcher, which will add start and stop events
to all dispatches.

- [x] Documentation
- [x] Tests

Contributes to https://github.com/commanded/commanded/issues/387

### DIFF
--- a/lib/application.ex
+++ b/lib/application.ex
@@ -1,4 +1,31 @@
 defmodule Commanded.Application do
+  use TelemetryRegistry
+
+  telemetry_event(%{
+    event: [:commanded, :application, :dispatch, :start],
+    description: "Emitted when an application starts dispatching a command",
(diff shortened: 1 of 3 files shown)


slashdotdash:
It would be possible to support `[:commanded, :application, :dispatch, :exception]` events if the aggregate returned the stacktrace on exception. This is currently being ignored by the [`ExecutionContext.format_reply/3` function](https://github.com/commanded/commanded/blob/e8e3760815745206cd6ad5683de6d96d584c4c61/lib/commanded/aggregates/execution_context.ex#L105) so the stack trace is not returned to the dispatcher process. 

Do you think it is worth adding that?

davydog187:
I do think it would be worth adding the exception event

PR #425, opened 2020-12-30
Mark ProcessManager callbacks as optional

davydog187:
- `init/1`
- `handle/2`
- `apply/2`
- `error/3`
- `interested?/1`

This will improve the documentation, see https://hexdocs.pm/elixir/GenServer.html#c:handle_info/2 for an example 

Screen Shot 2020-12-30 at 6 27 15 PM

### DIFF
--- a/lib/commanded/process_managers/process_manager.ex
+++ b/lib/commanded/process_managers/process_manager.ex
@@ -370,6 +370,8 @@ defmodule Commanded.ProcessManagers.ProcessManager do
               | {:skip, :continue_pending}
               | {:stop, reason :: term()}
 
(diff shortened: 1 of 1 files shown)


lib/commanded/process_managers/process_manager.ex
```diff
@@ -370,6 +370,8 @@ defmodule Commanded.ProcessManagers.ProcessManager do
               | {:skip, :continue_pending}
               | {:stop, reason :: term()}
 
+  @optional_callbacks init: 1, apply: 2
```
slashdotdash, line 373: The `interested?/1`, `handle/2`, and `error/3` functions are also optional.

PR #428, opened 2021-01-05
Have JsonSerializer implement EventStore.Serializer

davydog187:
This has a minor benefit, but if we were to modify the EventStore.Serializer behaviour, this would at least provide a warning. For now it serves as documentation to those reading the code (like myself).


davydog187:
I couldn't figure out why dialyzer was breaking, until I remembered that the `EventStore` is an optional dependency that's configurable.

I guess we can't move forward with this, unfortunately.

PR #429, opened 2021-01-25
Retry command execution on node down

drteeth:
Retry failed command dispatch due to the registered node for an aggregate being down when we try to execute against it.

Given an Aggregate A that was spawned on Node X and a command for A is
received on Node Y. If X is shutdown/killed after Y has
located the process for A on X, then the command will fail.

This PR allows the command to be retried, which will result in a new process
for A being opened on Node Y and the command will succeed from the
caller's perspective.


slashdotdash:
Thanks @drteeth.

PR #438, opened 2021-04-13
EventStore telemetry events

davydog187:
Adds telemetry events to the `Commanded.EventStore` adapter

- [x] ~~Determine API coverage that we want (I'm starting with the functions that the aggregate uses)~~
- [x] Tests
- [x] Pass relevant metadata
- [x] Documentation

I am going to avoid using the TelemetryRegistry for documenmting these events until it has better support for generating span documentation, otherwise, it will overtake this module. If you're ok with that @slashdotdash then I'm happy to convert to `telemetry_event` definitions

### API To cover
- [x] ack_event/3
- [ ] adapter/2
- [x] append_to_stream/4
- [x] delete_snapshot/2
- [x] delete_subscription/3
- [x] read_snapshot/2
- [x] record_snapshot/2
- [x] stream_forward/2
- [x] stream_forward/3
- [x] stream_forward/4
- [x] subscribe/2
- [x] subscribe_to/5
- [x] subscribe_to/6
- [x] unsubscribe/2

### DIFF
--- a/lib/commanded/event_store.ex
+++ b/lib/commanded/event_store.ex
@@ -13,14 +13,22 @@ defmodule Commanded.EventStore do
   Append one or more events to a stream atomically.
   """
   def append_to_stream(application, stream_uuid, expected_version, events) do
-    {adapter, adapter_meta} = Application.event_store_adapter(application)
+    meta = %{
+      application: application,
+      stream_uuid: stream_uuid,
+      expected_version: expected_version
+    }
 
-    adapter.append_to_stream(
-      adapter_meta,
-      stream_uuid,
-      expected_version,
-      events
-    )
+    span(:append_to_stream, meta, fn ->
+      {adapter, adapter_meta} = Application.event_store_adapter(application)
+
+      adapter.append_to_stream(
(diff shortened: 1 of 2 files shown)


davydog187:
Testing here https://github.com/SimpleBet/opentelemetry_commanded/pull/5

lib/commanded/event_store.ex
```diff
@@ -213,4 +225,10 @@ defmodule Commanded.EventStore do
 
     {adapter, config}
   end
+
+  defp span(event, meta, func) do
+    :telemetry.span([:commanded, :event_store, event], meta, fn ->
+      {func.(), %{}}
```
davydog187, line 231:
```suggestion
      {func.(), meta}
```

davydog187:
@slashdotdash I've updated the comment to reflect the telemetry coverage for the `Commanded.EventStore` module. I'm in the middle of writing tests and documentation. Can you let me know if the above coverage makes sense, or if I should also cover some of the other APIs

slashdotdash:
Thank you for the PR Dave!

PR #442, opened 2021-05-11
chore: add @impl to DefaultLifespan

brunohkbx:
Elixir introduced `@impl` to aid in the correct implementation of behaviours


slashdotdash:
Thanks @brunohkbx.

PR #447, opened 2021-06-04
Small documentation change

cdegroot:
This one tripped me up - we had 4 event handlers and it took me a while before I remembered that the default for `:max_restarts` is really low :)


slashdotdash:
Thanks @cdegroot!

Issue #445, opened 2021-05-14
Configurable UUID generator


slashdotdash:
Commanded uses the `elixir_uuid` library and `UUID.uuid4()` to create version 4 randomly generated UUIDs. This uses pseudo-random bytes generated by the `crypto` module. These UUIDs are used for dispatched command identity (`command_uuid`) and correlation id (`correlation_id`) and within the Commanded test suite.

It would be useful to allow the UUID generator to be configurable for a Commanded application and make the `elixir_uuid` library an optional dependency. The Commanded Application config could be extended to specify the UUID generator as a function.

For performance reasons it can be preferable to generate UUIDs which are not entirely random. One example would be the `event_id` for events stored in the Postgres database by the EventStore library which is the primary key. A database index on random values does not perform well due to writing to random positions in the index. Allowing partially sequential UUIDs to be generated should improve performance. To do this the `Commanded.EventStore.EventData` struct needs to be extended to include an `event_id` field which would allow Commanded to generate a UUID for each event using the configured UUID generator. The event store adapters then need to be updated to use this new field when persisting events. 

### Example

```elixir
defmodule MyApp.Application do
  use Commanded.Application,
    otp_app: :my_app,
    event_store: [
      adapter: Commanded.EventStore.Adapters.EventStore,
      event_store: MyApp.EventStore
    ],
    uuid_generator: &MyUUID.new/1

  router(MyApp.Router)
end
```

PR #450, opened 2021-07-03
Allow uuid generator to be configurable

FrancisMurillo:
Allow uuid generator to be configurable via `Commanded.Application` `uuid_generator` options. (#445)

One caveat is with dynamic apps. During `Commanded.Middleware.before_dispatch`, the `command_uuid` and `correlation_id` is empty which may break some client assumption. For compatibility, it is set after `before_dispatch`.

I do need some guidance on how to update the guide specially this new option and the ecosystem of adapters. Also since `elixir_uuid` is optional, the docs might have to specify that as well.


yordis:
I would suggest using https://github.com/bitwalker/uniq and https://github.com/bitwalker/uniq_compat and try to remove `elixir_uuid` dependency altogether. And maybe we wouldn't need `lib/commanded/uuid.ex` as well.

lib/application.ex
```diff
@@ -387,4 +390,32 @@ defmodule Commanded.Application do
   @doc false
   @spec registry_adapter(Commanded.Application.t()) :: {module, map}
   def registry_adapter(application), do: Config.get(application, :registry)
+
+  @type uuid :: String.t()
+
+  @doc false
+  @spec uuid_generator(Commanded.Application.t(), boolean()) :: (() -> uuid) | nil
+  def uuid_generator(application, true) do
```
Freyskeyd, line 398: I would have call this method `uuid_generator!` and removing the last argument

```suggestion
  def uuid_generator!(application) do
```

lib/commanded/assertions/event_assertions.ex
```diff
@@ -191,7 +191,8 @@ defmodule Commanded.Assertions.EventAssertions do
   @doc false
   def with_subscription(application, callback_fn, opts \\ [])
       when is_function(callback_fn, 1) do
-    subscription_name = UUID.uuid4()
+    uuid_generator = Commanded.Application.uuid_generator(application, true)
```
Freyskeyd, line 194: You can maybe use only one line here? we don't need `uuid_generator` afterward

lib/commanded/commands/dispatcher.ex
```diff
@@ -50,6 +51,18 @@ defmodule Commanded.Commands.Dispatcher do
...
+    uuid_generator = Application.uuid_generator(pipeline.application, true)
+
+    uuid_updater = fn
+      nil -> uuid_generator.()
+      value -> value
+    end
+
+    pipeline =
+      pipeline
+      |> Map.update!(:command_uuid, uuid_updater)
```
Freyskeyd, line 63: here you could have use [Map.put_new_lazy](https://hexdocs.pm/elixir/1.12/Map.html#put_new_lazy/3)

lib/commanded/commands/router.ex
```diff
@@ -507,16 +507,18 @@ defmodule Commanded.Commands.Router do
...
 
         defp do_dispatch(%@command_module{} = command, opts) do
+          alias Commanded.Application
           alias Commanded.Commands.Dispatcher
           alias Commanded.Commands.Dispatcher.Payload
 
           opts = Keyword.merge(@command_opts, opts)
 
           application = Keyword.fetch!(opts, :application)
+          uuid_generator = Application.uuid_generator(application, false) || fn -> nil end
```
Freyskeyd, line 517: Maybe `uuid_generator` should return an `fn -> nil end` instead of just `nil` ?

Freyskeyd (changes requested):
Hello and thank's for this amazing PR !

I've made some suggestions but I'm also concern about the `UUID` module which seems to be unnecessary as pointed by @yordis 

Maybe can we completely remove it as suggested?

lib/commanded/commands/dispatcher.ex
FrancisMurillo: Using `Map.put_new_lazy` does not work on structs since the key is always present and will be a NOOP.

lib/commanded/assertions/event_assertions.ex
FrancisMurillo: Okay. Made it more inline

lib/application.ex
FrancisMurillo: Good suggestion, applied.

lib/commanded/commands/router.ex
FrancisMurillo: Also updated  `Application.uuid_generator!` returning an a zero arity function

FrancisMurillo:
So should have I retained `Commanded.UUID` and remove `uniq_compat`?

slashdotdash:
> So should have I retained `Commanded.UUID` and remove `uniq_compat`?

Yes, I think so as dropping the dependency removes the potential for version clashes.

lib/commanded/commands/dispatcher.ex
Freyskeyd: Oh you right I forgot about the struct!

FrancisMurillo:
Restored `Commanded.UUID` while applying the other comments

edwardzhou:
There are two PRs for the same purpose. May I know which one is better to merge?
@slashdotdash

cdegroot:
I take it that #474 overrides this and we can close this?

yordis:
I would close it until further notice

jwilger:
Closing as per discussion. #474 was closed, also, but due to #493 being merged.

PR #451, opened 2021-07-14
Incorrect aggregate_version when linking events

boatkorachal:
When linking the events from one stream to another stream following this guide https://github.com/commanded/recipes/issues/16, the aggregate in another stream shows wrong expected version when dispatch a command because it uses original_stream_version instead of stream_version.

### DIFF
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -407,7 +407,7 @@ defmodule Commanded.Aggregates.Aggregate do
   # Handle events appended to the aggregate's stream, received by its
   # event store subscription, by applying any missed events to its state.
   defp handle_event(%RecordedEvent{} = event, %Aggregate{} = state) do
-    %RecordedEvent{data: data, stream_version: stream_version} = event
+    %RecordedEvent{data: data, event_number: event_number} = event
 
     %Aggregate{
       aggregate_module: aggregate_module,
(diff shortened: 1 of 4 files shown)


drteeth:
@boatkorachal sorry to get to after such a long time, is this still a problem for you? If so, can you explain the problem a bit more?

Are you trying to use an EventHandler do link events from AggregateA to the stream of AggregateB ? (Another aggregate)

boatkorachal:
Hi @drteeth, sorry, I’m no longer working on the project, but I believe the issue might still persist. The `stream_version` in the aggregate's state is getting from the `stream_version` in RecordedEvent, which mapped from the `original_stream_version` in EventStore. So within a stream, `original_stream_version` doesn't guarantee incremental ordering. 

For example, consider two streams:

- `stream_1` has `stream_version=10` and `original_stream_version=10`
- `stream_2` has `stream_version=20` and `original_stream_version=20`

If I link the last event from `stream_1` to `stream_2`, the last event in `stream_2` will have `stream_version=21` and `original_stream_version=10`. As a result, the aggregate for stream_2, which currently has `aggregate_version=20`, will try to consume `%RecordedEvent{event_number: 21, stream_version: 10}` and skip (or fail depending on `stream_version` number). So I think if we update `aggregate_version` from `event_number` in `RecordedEvent`, instead of `stream_version`, could solve the issue.

jwilger:
Closing this for now. If someone actively working with commanded is running into a similar issue and wants to take over this PR, feel free to reopen it.

PR #456, opened 2021-09-07
Allow for telemetry 1.0

michaelherold:
The bump to 1.0 only solidifies the public API. It's the same as the 0.4 series, just with a guarantee of a stable API. Since there are libraries that have not updated yet, accepting both for now seems reasonable.


slashdotdash:
Thanks for the pull request Michael.

Issue #441, opened 2021-05-06
Allow a process manager to stop after dispatching a command


slashdotdash:
A process manager instance can only be stopped by returning `:stop` from the `interested?/1` callback function. This doesn't allow you to decide when to stop the instance based upon its state and/or a recently dispatched command. It would be useful to support stopping an instance from either the `handle/2` function or as a new callback function, such as `after_command/1`.

### Add an `after_command/1` callback function

The return value from the callback function would control when the process manager instance is terminated. The callback function would only be called after the command is successfully dispatched. A dispatch error would call the existing `error/3` function.

#### Example

```elixir
defmodule ExampleProcessManager do
  use Commanded.ProcessManagers.ProcessManager,
    application: ExampleApp,
    name: "ExampleProcessManager"

  def handle(%ExampleProcessManager{}, %AnEvent{}) do
    [%SomeCommand{}]
  end

  def after_command(%ExampleProcessManager{}, %SomeCommand{}) do
    :stop
  end
end
```

This is similar to the existing `Commanded.Aggregates.AggregateLifespan` behaviour which contains an `after_command/1` function.

### Allow a process manager `handle/2` function to stop an instance

Allow a process manager's `handle/2` callback function to return a tuple in the format `{command, :stop}` to dispatch one or more commands and then stop the process manager instance.

#### Example

```elixir
defmodule ExampleProcessManager do
  use Commanded.ProcessManagers.ProcessManager,
    application: ExampleApp,
    name: "ExampleProcessManager"

  def handle(%ExampleProcessManager{}, %AnEvent{}) do
    command = %SomeCommand{}

    {command, :stop}
  end
end
```

philipgiuliani:
Are there any plans to implement the second version (`{command(s), :stop}`). The problem is that I emit multiple commands multiple times, but in a special `handle` clause I need to stop it after emitting the commands. The only workaround I can think of right now, is to create a custom command which I will emit, just to stop the process manager afterwards.

PR #460, opened 2021-11-02
Allow a process manager to stop after dispatching a command

lucazulian:
fixes https://github.com/commanded/commanded/issues/441

This PR allow a process manager instance to be stopped through `after_command/1` callback function.
The`after_command/1` callback function can have two return values, `:stop` that would stop with specific command and `:continue` that would ignore the command.
The callback function would only be called after the command is successfully dispatched. 
After commands are successfully dispatched the Process state snapshot would be deleted as it would be in the `interested?/1` callback implementation.

### DIFF
--- a/lib/commanded/process_managers/process_manager_instance.ex
+++ b/lib/commanded/process_managers/process_manager_instance.ex
@@ -184,8 +184,6 @@ defmodule Commanded.ProcessManagers.ProcessManagerInstance do
     %RecordedEvent{correlation_id: correlation_id, event_id: event_id, event_number: event_number} =
       event
 
-    %State{idle_timeout: idle_timeout} = state
-
     telemetry_metadata = telemetry_metadata(event, state)
     start_time = telemetry_start(telemetry_metadata)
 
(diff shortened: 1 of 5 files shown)


lib/commanded/process_managers/process_manager_instance.ex
```diff
@@ -350,6 +348,33 @@ defmodule Commanded.ProcessManagers.ProcessManagerInstance do
...
+      process_state: process_state
+    } = state
+
+    case process_manager_module.after_command(process_state, command) do
+      :stop ->
+        Logger.debug(fn ->
+          describe(state) <> " has been stopped by command " <> inspect(command)
+        end)
+
+        :ok = delete_state(state)
```
lucazulian, line 369: I don't know if it's correct to delete instance here, in the `c:interested?` it's done by `ProcessRouter` but in this case i don't see any way to handle by it

slashdotdash: Doing this is fine as the process router calls the process manager instance to delete its state.

lucazulian: Ohh thank you 🙏🏿

PR #463, opened 2021-11-22
Event Handler not calling init on restart

marpo60:
Hello!

I've experience the following issues that I don't think its the intended behaviour.

Im using and Event Handler state for saving dynamic information and restoring on the `init/1` callback.
Last week a EH crashed to my surprise the EH process, the new one that the supervisor started, did not run the `init/1`
The reason for this is that `init/1` was called once in the `child_spec` definition and it was not calling the module start link.
So what end up happening is that the init/1 is called only once at the first start of the Event Handler

With this change the only downside that im seeing is that `parse_config!` is called in the `child_spec` and also in the `start_link`

I haven't found a way to get around that, because in the `init/1` callback you can define the application and the name, so calling it in the child_spec is needed in the same way is needed to be called in the `start_link/1`

### DIFF
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -517,10 +517,8 @@ defmodule Commanded.Event.Handler do
       def child_spec(opts) do
         opts = Keyword.merge(@opts, opts)
 
-        {application, name, _config} = Handler.parse_config!(__MODULE__, opts)
-
         default = %{
-          id: {__MODULE__, application, name},
+          id: {__MODULE__, opts},
           start: {__MODULE__, :start_link, [opts]},
           restart: :permanent,
           type: :worker
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -517,11 +517,11 @@ defmodule Commanded.Event.Handler do
       def child_spec(opts) do
         opts = Keyword.merge(@opts, opts)
 
-        {application, name, config} = Handler.parse_config!(__MODULE__, opts)
+        {application, name, _config} = Handler.parse_config!(__MODULE__, opts)
 
         default = %{
           id: {__MODULE__, application, name},
-          start: {Handler, :start_link, [application, name, __MODULE__, config]},
+          start: {__MODULE__, :start_link, [opts]},
           restart: :permanent,
           type: :worker
         }
--- a/test/event/event_handler_state_test.exs
+++ b/test/event/event_handler_state_test.exs
@@ -77,7 +77,7 @@ defmodule Commanded.Event.EventHandlerStateTest do
       assert_receive {:event, ^event1, metadata}
       assert match?(%{state: 11}, metadata)
 
-      %{id: id} = StatefulEventHandler.child_spec([])
+      %{id: id} = StatefulEventHandler.child_spec([state: 10])
 
       stop_supervised!(id)
 
--- a/test/event/event_handler_state_test.exs
+++ b/test/event/event_handler_state_test.exs
@@ -77,7 +77,7 @@ defmodule Commanded.Event.EventHandlerStateTest do
       assert_receive {:event, ^event1, metadata}
       assert match?(%{state: 11}, metadata)
 
-      %{id: id} = StatefulEventHandler.child_spec([state: 10])
+      %{id: id} = StatefulEventHandler.child_spec(state: 10)
 
       stop_supervised!(id)
 
--- a/test/event/handler_init_test.exs
+++ b/test/event/handler_init_test.exs
@@ -34,6 +34,15 @@ defmodule Commanded.Event.HandlerInitTest do
       assert_handler_name(handler2, "Commanded.Event.RuntimeConfigHandler.tenant2")
       assert_handler_name(handler3, "Commanded.Event.RuntimeConfigHandler.tenant3")
     end
+
+    test "should be called on restart if the process crashes" do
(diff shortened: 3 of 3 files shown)


davydog187:
This bug caused to production issues for us this week :(

slashdotdash:
I agree that the current behaviour isn't what one would expect and should be fixed.

The `child_spec/1` function in the event handler module needs to call the optional `init/1` callback function since the identity of the child spec used to guarantee uniqueness by the supervisor depends upon the Commanded application and the event handler's name. These can both be provided at either compile-time or runtime. 

With the change you suggest the `init/1` callback function would get called by both `child_spec/1` and `start_link/1` which is also not ideal. This means the `init/1` callback would always be called n+1 times where n is the number of process (re)starts. If you added a `refute_receive {:init, :tenant1}` assertion to the new test in this PR then it would fail as `init/1` is being called three times when the process is restarted once.

Looking at how this problem is solved elsewhere, in the Ecto library they define an `init/2` function which is also passed an additional `context` argument which is either `:supervisor` or `:runtime`. 

```elixir
@doc """
  A callback executed when the repo starts or when configuration is read.

  The first argument is the context the callback is being invoked. If it
  is called because the Repo supervisor is starting, it will be `:supervisor`.
  It will be `:runtime` if it is called for reading configuration without
  actually starting a process.

  The second argument is the repository configuration as stored in the
  application environment. It must return `{:ok, keyword}` with the updated
  list of configuration or `:ignore` (only in the `:supervisor` case).
  """
@callback init(context :: :supervisor | :runtime, config :: Keyword.t()) ::
              {:ok, Keyword.t()} | :ignore
```

A similar approach could be used in Commanded where  an extra `context` flag added to the `config` could be used to determine whether the calling context is the supervisor's `child_spec/1` or starting the actual process.

```elixir
def init(config) do
  case Keyword.fetch!(config, :context) do
    :start_link -> 
      # called when starting the process, can populate state here
      {:ok, config}

    :child_spec -> 
      # called during `child_spec/1` and must return application and handler name
      {:ok, config}
  end
end
```

Any thoughts?

marpo60:
Hi @slashdotdash.

Thanks for the thoughtful response.

> With the change you suggest the init/1 callback function would get called by both child_spec/1 and start_link/1 which is an improvement, but not ideal.

You are correct on this.

I thought about your proposed idea, and although I understand the benefits of adding a new context variable and based on that do one thing or the other.

For the users that upgrade the library and do not use this new context variable, will be potentially doing the expensive init n+1 times.

I've though about it a little bit more and seems like: 

```
     def child_spec(opts) do
        opts = Keyword.merge(@opts, opts)

        {application, name, config} = Handler.parse_config!(__MODULE__, opts)

        default = %{
          id: {__MODULE__, application, name},
          start: {Handler, :start_link, [application, name, __MODULE__, config]},
          restart: :permanent,
          type: :worker
        }

        Supervisor.child_spec(default, [])
      end
```

And the cause of all is that the same function `Handler.parse_config!/2` is doing 2 things, fetching static information for the application and name and also doing the init which might include state.

So the question, become how do we split the `Handler.parse_config/2` and have one method for the `child_spec` and other for `start_link`

For process manager the way of getting the child_spec id is based on in the return of the `interested?/1`

Would it be possible to have a new function `defoverridable parse_config: 1`

And you could define in the EH your `parse_config/1` to at runtime return application and name. Avoid calling init in the child spec and in start_link call both

```
      def start_link(opts \\ []) do
        opts = Keyword.merge(@opts, opts)

        {application, name} = __MODULE__.parse_config(opts)
        {:ok, config} == __MODULE__.init(opts)

        Handler.start_link(application, name, __MODULE__, config)
      end
```

```
      def child_spec(opts) do
        opts = Keyword.merge(@opts, opts)

        {application, name} = __MODULE__.parse_config(opts)

        default = %{
          id: {__MODULE__, application, name},
          start: {__MODULE__, :start_link, [opts]},
          restart: :permanent,
          type: :worker
        }

        Supervisor.child_spec(default, [])
      end
```

Thoughts?

slashdotdash:
I had some more thoughts on this over the weekend. The purpose of calling the `parse_config/1` (and hence also call the `init/1` callback function) in the `child_spec/1` function is to get the Commanded application and event handler name which is then used as in the child spec `id`. The purpose of this is to ensure only one child is started by the supervisor for a given Commanded application and event handler name. It might be sufficient to use the provided `opts` Keyword as part of the child spec `id` along with the event handler module to guarantee uniqueness.

```elixir
def child_spec(opts) do
  opts = Keyword.merge(@opts, opts)

  default = %{
    id: {__MODULE__, opts},
    start: {Handler, :start_link, [opts]},
    restart: :permanent,
    type: :worker
  }

  Supervisor.child_spec(default, [])
end
```

Doing the above would mean that the `init/1` callback could then be called within the `start_link/1` function each time the process is started. As we would prefer.

marpo60:
Did the change.

By doing this change we are introducing a new uniqueness that is not only with `application` and `name`

Had to fix a failing test because now all the keys in opts are part of the id of the child_spec

```diff
-      %{id: id} = StatefulEventHandler.child_spec([])
+      %{id: id} = StatefulEventHandler.child_spec([state: 10])
```

This means that now  `StatefulEventHandler.child_spec([])` is different than  `StatefulEventHandler.child_spec([any: :thing])`

@slashdotdash do you think this is an issue?

slashdotdash:
@marpo60 I'm happy with these changes, so will merge and release a v1.3.1 hotfix.

PR #468, opened 2022-01-04
Use Elixir 1.13 locally and in CI

davydog187:


.github/workflows/test.yml
```diff
@@ -8,8 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        otp: ['23.3']
-        elixir: ['1.12.2']
+        otp: ['23.3', '24.2']
+        elixir: ['1.12.2', '1.13.1']
```
davydog187, line 12: I'm only going to run on the latest, i.e. removing 1.12. I'm not sure I need both otp version either

slashdotdash:
Merged in 5e4f36fe813d68dcfaa74f7ac10b9733a7bcbe49.

PR #471, opened 2022-01-27
Support batched events to Event.Handler

### LINKED ISSUE #261: Event handler support for receiving events in batches
Event handlers receive events one at a time for processing. Returning `:ok` indicates successful handling, the event is acknowledged, then the next event can be received.

```elixir
defmodule ExampleHandler do
  use Commanded.Event.Handler, name: __MODULE__

  def handle(event, _metadata) do
    # Process single event
    :ok
  end
end
```

## Proposal

Add support to event handlers allowing configuration of the batch size of events to receive.  

```elixir
defmodule ExampleHandler do
  use Commanded.Event.Handler, name: __MODULE__, batch_size: 50

  def handle(events, _metadata) when is_list(events) do
    # Process list of events
    :ok
  end
end
```

This would be the maximum number of events that would be processed if there is a backlog of events. If the handler is up to date then events will be sent immediately, it won't wait for an entire batch size worth of events.

davydog187:
This PR intends to add support for reaching batches of events from event stores and their respective adapters.

Closes #261 

- [ ] Solidify the API and contract for metadata and errors
- [ ] Document the API with examples
- [ ] Complete test coverage

### DIFF
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -376,6 +376,17 @@ defmodule Commanded.Event.Handler do
               | {:error, reason :: any()}
 
   @doc """
+  Handle a batch of domain events and their metadata.
+
+  TODO describe API
+  """
+  @callback handle_batch([domain_event], metadata) ::
+              :ok
+              | {:ok, new_state :: any()}
+              | {:error, :already_seen_event}
+              | {:error, reason :: any()}
+
+  @doc """
   Called when an event `handle/2` callback returns an error.
 
   The `c:error/3` function allows you to control how event handling failures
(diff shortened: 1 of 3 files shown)


davydog187:
Meant to open this as a draft. Regardless, working on it!

lib/commanded/event/handler.ex
```diff
@@ -459,7 +470,7 @@ defmodule Commanded.Event.Handler do
               | :skip
               | {:stop, reason :: term()}
 
-  @optional_callbacks init: 0, init: 1, error: 3
+  @optional_callbacks init: 0, init: 1, error: 3, handle: 2, handle_batch: 2
```
davydog187, line 473: users should implement either or. We could throw an error at compile time in the `__using__` macro if we want to

slashdotdash: I agree it would be useful to throw an error if an event handler module implements both `handle/2` and `handle_batch/2` callback functions.

davydog187:
Going to close this as I haven't had the time to come back to it 😢

PR #474, opened 2022-02-14
chore: replace elixir_uuid pkg

yordis:


edwardzhou:
Good job 👍  It will be great if use uuid6 instead of uuid4.

davydog187:
@yordis can you please include the motivation for the change?

yordis:
https://github.com/zyro/elixir-uuid/issues/50

Screen Shot 2022-02-23 at 11 12 36 AM

davydog187:
I'm sorry @yordis but I'm still not seeing the justification. Is there a missing feature that you're looking for?

yordis:
https://github.com/commanded/commanded/pull/474#issuecomment-1039061214
https://github.com/commanded/commanded/pull/474#issuecomment-1048954921

davydog187:
I'm not sure why you're being so terse @yordis, but I'll write out the justification as I understand it.

The current package that commanded depends on, [elixir_uuid](https://github.com/zyro/elixir-uuid) has not been updated since 2019. In an effort to support `uuid6`, this PR is adding [uniq](https://github.com/bitwalker/uniq) by Bitwalker to create a path towards that support.

yordis:
@davydog187 correct, we spoke in the Slack channel about it

> I'm not sure why you're being so terse @yordis

Hammer today with work where I just finish the last mile and am ready to disconnect, my apologies if came across as tense, just trying to respond between my means.

yordis:
Can we do it incrementally?

slashdotdash:
> Can we do it incrementally?

Yes, the two steps above could be done as separate PRs.

yordis:
@slashdotdash alright, let me do the replacement

bitwalker:
My recommended approach would be to implement the abstractions needed for #450, but instead of implementing UUID within Commanded, just declare the dependency on `elixir_uuid` optional, so that alternatives can be provided without forcing users to fetch a dependency they aren't actually using.

slashdotdash:
Making the `:elixir_uuid` package an optional dependency would require adding an additional step to add the library for users when getting started with Commanded. It will also require upgrading users to manually add the dependency. I try to minimise such manual setup and minimise manual upgrade work wherever possible. That was the reasoning for copying the UUID functionality into Commanded as it will be backwards compatible and won't require any changes during an upgrade of Commanded.

edwardzhou:
@yordis I have submitted a PR to your branch, to synchronize with the main stream and resolve conflicts.

dvic:
A few days ago @slashdotdash merged my PR in eventstore: https://github.com/commanded/eventstore/pull/253

Maybe we should go with the same route here? (vendoring the UUID module)

slashdotdash:
Yes, that’s the approach I would prefer to take.

slashdotdash:
I've created pull request #493 to remove the `elixir_uuid` dependency.

slashdotdash:
Superceeded by #493.

PR #477, opened 2022-02-25
Add credo code analysis

aleDsz:
It's just a simple PR to implement an useful tool to preserve the good code quality

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -1,0 +1,14 @@
+%{
+  configs: [
+    %{
+      name: "default",
+      files: %{
+        included: ["lib/"],
+        excluded: [~r"/_build/", ~r"/deps/", ~r"/node_modules/", ~r"/test/"]
+      },
+      strict: true,
+      parse_timeout: 5000,
+      color: true,
+    }
+  ]
+}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -28,10 +28,16 @@ jobs:
         restore-keys: ${{ runner.os }}-mix-
 
     - name: Install dependencies
-      run: mix deps.get
+      run: mix do deps.get, deps.compile
+
+    - name: Check compilation
+      run: mix compile --warnings-as-errors
 
     - name: Check formatting
       run: mix format --check-formatted
+
+    - name: Analyse code
+      run: mix credo
 
     - name: Run tests
       run: |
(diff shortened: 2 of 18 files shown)


slashdotdash:
@aleDsz I'm not keen on making such large changes to the codebase based upon Credo recommendations. My concern is introducing bugs or subtle changes in behaviour during refactoring.

aleDsz:
@slashdotdash If I split all these changes into little PRs, would be a better way to all this changes be merged?

slashdotdash:
I'm not sure what value the proposed changes add. For example they include adding many new functions which only have a single line in the function body. I'd prefer not to need to strictly follow conventions dictated by a tool such as Credo when they don't make they code easier to reason about or follow.

aleDsz:
> I'm not sure what value the proposed changes add. For example they include adding many new functions which only have a single line in the function body. I'd prefer not to need to strictly follow conventions dictated by a tool such as Credo when they don't make they code easier to reason about or follow.

The proposal is to standardize the entire style guide used in the project, so that there are no parts of the code that are so different from the rest of the project.

The idea is not to bring meaningless modifications, since some functions are very complex, I had to change them so that the public functions were just "endpoints" of a flow of private functions to perform the pattern matching of existent `case... do` blocks.

The "value" of these changes would be to make it easier for those who want to contribute to the project (like me) and for them to write code that is still in the style in which the project currently operates.

If it is not of interest to everyone, I will understand and will not open PR to implement the use of `credo`.

Now, about `credo`, we can disable certain checks made by `credo` so that the project can create its own style of code and check only what everyone understands to be paramount and remove those that don't make much sense to the project. Or, we can change how complex a function should be and things like that.

slashdotdash:
My preference is to follow the existing coding style and strive to continuously improve the code when making changes. The "leave it better than you found it" principle.

slashdotdash:
If you feel Credo really adds value then submit a PR where it is added as an optional Mix dependency only. Don't include it in the GitHub Actions workflow, nor include any suggested code refactoring. Then I can include the appropriate rules as required.

aleDsz:
> If you feel Credo really adds value then submit a PR where it is added as an optional Mix dependency only. Don't include it in the GitHub Actions workflow, nor include any suggested code refactoring. Then I can include the appropriate rules as required.

Sure

PR #479, opened 2022-03-02
Add credo dependency as optional

aleDsz:
As we talked at #477, i'm adding the `credo` dependency as optional so you can evaluate what's better for the existing coding style


mix.exs
```diff
@@ -65,6 +65,7 @@ defmodule Commanded.Mixfile do
       # Optional dependencies
       {:jason, "~> 1.2", optional: true},
       {:phoenix_pubsub, "~> 2.0", optional: true},
+      {:credo, "~> 1.6", optional: true},
```
yordis, line 68:
```suggestion
      {:credo, ""~> 1.6"", only: [:dev, :test], runtime: false}
```

yordis (changes requested):

PR #484, opened 2022-05-06
chore: improve typespec

yordis:


lib/application.ex
```diff
@@ -205,6 +207,9 @@ defmodule Commanded.Application do
         Supervisor.stop(pid, :normal, timeout)
       end
 
+      @doc """
+      TODO: Does this function need to be documented?
+      """
```
yordis, line 212: @slashdotdash what was the intention behind this function? I see it in my documentation behind my App module so I would like to improve the typespec and whatnot because of it.

slashdotdash: The documentation can be copied from the similar function defined in [`lib/commanded.ex`](https://github.com/commanded/commanded/blob/master/lib/commanded.ex#L43).

lib/application.ex
```diff
@@ -165,6 +165,7 @@ defmodule Commanded.Application do
   """
 
   @type t :: module
+  @type options :: [[name: nil | atom] | Supervisor.init_option()]
```
slashdotdash, line 168: Only the `:name` option is currently supported by the [Commanded application supervisor](https://github.com/commanded/commanded/blob/master/lib/commanded/application/supervisor.ex#L39). The supervisor init options (`:strategy`, `:max_restarts`, and `:max_seconds`) aren't supported.

```suggestion
  @type options :: [{:name, atom() | nil}]
```

lib/application.ex
```diff
@@ -187,6 +188,7 @@ defmodule Commanded.Application do
         config
       end
 
+      @spec child_spec(opts :: Commanded.options()) :: Supervisor.child_spec()
```
slashdotdash, line 191: For consistency with the `start_link/1` callback function typespec:

```suggestion
      @spec child_spec(opts :: options()) :: Supervisor.child_spec()
```

slashdotdash (changes requested):

lib/application.ex
yordis: I fixed it, 

I am also a bit confused, https://github.com/commanded/commanded/blob/9b7e89416f1f9a9bff20e458cb445768dbce30c9/lib/commanded/application/supervisor.ex#L39-L45

You pass the entire `init_option` down based on that code, no?

If you don't mind to help me to clarify myself.

slashdotdash: It looks as though `:name` is the only option supported by the [`Supervisor.start_link/3` function](https://hexdocs.pm/elixir/Supervisor.html#start_link/3) and currently only the [`:strategy` option is passed to the `Supervisor.init/2` function](https://github.com/commanded/commanded/blob/9b7e89416f1f9a9bff20e458cb445768dbce30c9/lib/commanded/application/supervisor.ex#L62). We could extend it to support the additional `:max_restarts` and `:max_seconds` options if required.

slashdotdash:
Thanks for the pull request Yordis.

PR #486, opened 2022-06-19
Event handler concurrency

### LINKED ISSUE #184: Event handler concurrency
Allow event handlers to specify how many processes are started to allow handling events concurrently.

The behaviour is similar to Greg's Event Store which uses the [competing consumer](https://eventstore.org/docs/introduction/competing-consumers/) pattern. Increasing the concurrency means the event ordering guarantee is lost since events will be handled out-of-order. To alleviate this issue an optional partitioning function may be provided to ensure events are processed in order where necessary (e.g. per aggregate identity).

```elixir
defmodule ExampleHandler do
  use Commanded.Event.Handler, 
    name: "ExampleHandler", 
    concurrency: 10, 
    partition_by: fn _event, metadata -> metadata.stream_id end

  def handle(_event, _metadata) do
    :ok
  end
end
```

This feature depends upon Subscription concurrency ([#124](https://github.com/commanded/eventstore/issues/124)) to be implemented in the Elixir Event Store. Greg's Event Store already has support for concurrency.

slashdotdash:
Allow event handlers to specify  a `:concurrency` option to configure how many processes are started. This is to allow handling of events concurrently. The default behaviour is to start a single event handler process. 

Events will be distributed to event handler processes using a round-robin strategy for even load.

Increasing the concurrency means the event ordering guarantee is lost since events will be handled out-of-order. To alleviate this issue an optional `partition_by/2` callback function may be provided to configure which partition an event belongs to. This can be used to ensure events are processed in order where necessary (e.g. per aggregate stream).

```elixir
defmodule ConcurrentHandler do
  use Commanded.Event.Handler, 
    name: __MODULE__, 
    concurrency: 10

  @impl Commanded.Event.Handler
  def handle(event, metadata) do
    :ok
  end
 
  # Partition by stream (`stream_id` field from metadata map) so events 
  # from the same stream are handled in order.
  @impl Commanded.Event.Handler
  def partition_by(event, metadata) do
    %{stream_id: stream_id} = metadata

    stream_id
  end
end
```

Closes #184.


qhwa:
I just wanted to let you know that this is useful to us. Can't wait to use it.
Thank you for the excellent work!

PR #495, opened 2022-08-09
Fix EventData typespec

dvic:
I believe `causation_id` and `correlation_id` are optional fields (for example `lib/commanded/event/mapper.ex:55` sets to a value that can be nil), updated the typespec accordingly.

### DIFF
--- a/lib/commanded/event_store/event_data.ex
+++ b/lib/commanded/event_store/event_data.ex
@@ -8,7 +8,7 @@ defmodule Commanded.EventStore.EventData do
 
   @type t :: %Commanded.EventStore.EventData{
           causation_id: uuid() | nil,
-          correlation_id: uuid() | nil,
+          correlation_id: uuid(),
           event_type: String.t(),
           data: struct(),
           metadata: map()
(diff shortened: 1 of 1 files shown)


jvantuyl:
FYI, it looks like at least `correlation_id` will be automatically created if not specified.  See [here](https://github.com/commanded/commanded/blob/master/lib/commanded/commands/router.ex#L515).

lib/commanded/event_store/event_data.ex
```diff
@@ -7,8 +7,8 @@ defmodule Commanded.EventStore.EventData do
   @type uuid :: String.t()
 
   @type t :: %Commanded.EventStore.EventData{
-          causation_id: uuid(),
-          correlation_id: uuid(),
+          causation_id: uuid() | nil,
+          correlation_id: uuid() | nil,
```
jvantuyl, line 11:
```suggestion
          correlation_id: uuid(),
```
FYI, it looks like `correlation_id` will be automatically generated if not specified. See [here](https://github.com/commanded/commanded/blob/master/lib/commanded/commands/router.ex#L515).

drteeth:
Hi @dvic, if you accept @jvantuyl's suggestion, I'll see about getting this merged for you.

cdegroot (approved):

drteeth (approved):

PR #496, opened 2022-08-11
Allow to tweak batch size when reading events for an aggregate

cdegroot:
Make the batch size tweakable and set the default to the same default as in event_store.ex for consistency.


cdegroot:
Note: currently would not transport batching settings from app config. I'm torn with choosing between putting it inside snapshotting and having more explicit options to transport around. Opinions?

cdegroot:
Note: not super important, this was for a while a pressing issue for the enormous numbers of events we processed at $prevJob, but at the moment don't have an interest in this so I'm fine with closing this PR.

yordis:
Personally, I would change the name to just `batch_size` or without that weird `populate` name.

But, I haven't run into any stream that required me to read 500 events

jwilger:
In the name of just getting the PR and issue backlogs cleaned up quickly, so that we can concentrate more on current/upcoming PRs and issues, I'm going to close this as per @cdegroot's comment in order to be conservative and avoid adding to the public API. If anyone is interested in reviving this and addressing existing feedback, feel free to update and reopen. Test coverage should be added/updated, too.

PR #497, opened 2022-08-11
Set read event batch size to a somewhat saner value

cdegroot:
100 is fairly small, and 1_000 is used elsewhere (event_store.ex) as a default.


dvic:
Maybe make this also configurable?

```elixir
@read_event_batch_size Application.compile_env(:commanded, :aggregate_read_event_batch_size, 1000)
```

dvic:
Nvm didn't see #496

PR #503, opened 2022-09-05
Fix typespec typo in Commanded.Application

dvic:
I think dialyzer did not catch this one because it ignores test files and that's the only place this macro is called? This was introduced in #484


slashdotdash:
Thanks for fixing this Damir.

slashdotdash:
Published to Hex as [v1.4.1](https://hex.pm/packages/commanded/1.4.1)

Issue #336, opened 2020-01-15
Get aggregate state


slashdotdash:
An aggregate's current state can be accessed using the [`Commanded.Aggregates.Aggregate.aggregate_state/3` function](https://github.com/commanded/commanded/blob/3b8f08526b97a196e3f94e7c868cc059c9062652/lib/commanded/aggregates/aggregate.ex#L140). You pass the Commanded application, aggregate module, and aggregate identity:

```elixir
alias Commanded.Aggregates.Aggregate

%BankAccount{..} = Aggregate.aggregate_state(BankApp, BankAccount, account_number)
```

However this only works when the aggregate process has already been started by having a command dispatched to it. 

A helper function should be added to a Commanded application to allow an aggregate's state to be queried.

```elixir
{:ok, %BankAccount{}} = BankApp.aggregate_state(BankAccount, account_number)
```

Or using `Commanded.aggregate_state/3` where the application is provided as an argument:

```elixir
{:ok, %BankAccount{}} = Commanded.aggregate_state(BankApp, BankAccount, account_number)
```

Retrieving aggregate state should be done by querying the event store for an optional state snapshot and then replaying the aggregate's event stream. It is not necessary to start an aggregate `GenServer` process.

jwilger:
> A helper function should be added to a Commanded application to allow an aggregate's state to be queried. 

I'm curious about the motivation for that. Isn't directly querying an aggregate's state exactly what one is *not* supposed to do when using the CQRS pattern?

slashdotdash:
@jwilger Yes, it breaks the strict CQRS interpretation of defining a separate model for writes and reads. I would recommend having this separation for most use cases.

This is mostly designed to be a convenience helper function for those scenarios where read access patterns can be easily fulfilled by querying the aggregate state by its identity. An example from a recent project is where we use an isolated tenant-specific event store for each customer's data and didn't want or need to build a separate read model for certain aggregates. Accessing the aggregate state directly was good enough.

slashdotdash:
Closed by #448.

PR #505, opened 2022-09-08
feat: return nil aggregate state when the aggregate does not exist

fahchen:
I think it is more appropriate to return `nil` aggregate state when the aggregate does not exist. We can also build the initial aggregate state with `struct(aggregate_module)` easily.

### DIFF
--- a/lib/application.ex
+++ b/lib/application.ex
@@ -208,7 +208,8 @@ defmodule Commanded.Application do
       end
 
       @doc """
-      Retrieve aggregate state of an aggregate.
+      Retrieve aggregate state of an aggregate,
+      return `nil` if the aggregate does not exist.
 
       Retrieving aggregate state is done by calling to the opened aggregate,
       or querying the event store for an optional state snapshot
@@ -218,7 +219,7 @@ defmodule Commanded.Application do
               aggregate_module :: module(),
               aggregate_uuid :: Aggregate.uuid(),
               timeout :: integer
-            ) :: Aggregate.state()
+            ) :: Aggregate.state() | nil
       def aggregate_state(aggregate_module, aggregate_uuid, timeout \\ 5000) do
         Aggregate.aggregate_state(
           __MODULE__,
--- a/lib/commanded.ex
+++ b/lib/commanded.ex
@@ -34,7 +34,8 @@ defmodule Commanded do
   end
 
   @doc """
-  Retrieve aggregate state of an aggregate.
+  Retrieve aggregate state of an aggregate,
+  return `nil` if the aggregate does not exist.
 
   Retrieving aggregate state is done by calling to the opened aggregate,
   or querying the event store for an optional state snapshot
@@ -45,7 +46,7 @@ defmodule Commanded do
           aggregate_module :: module(),
           aggregate_uuid :: Aggregate.uuid(),
           timeout :: integer
-        ) :: Aggregate.state()
+        ) :: Aggregate.state() | nil
   def aggregate_state(application, aggregate_module, aggregate_uuid, timeout \\ 5_000) do
     Aggregate.aggregate_state(
       application,
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -225,7 +225,10 @@ defmodule Commanded.Aggregates.Aggregate do
               snapshotting: Snapshotting.new(application, aggregate_uuid, snapshot_options)
             }
             |> AggregateStateBuilder.populate()
-            |> Map.fetch!(:aggregate_state)
+            |> case do
+              %Aggregate{aggregate_version: 0} -> nil
+              %Aggregate{} = state -> Map.fetch!(state, :aggregate_state)
+            end
           end)
 
         case Task.yield(task, timeout) || Task.shutdown(task) do
--- a/test/aggregates/aggregate_state_test.exs
+++ b/test/aggregates/aggregate_state_test.exs
@@ -22,6 +22,19 @@ defmodule Commanded.Aggregates.AggregateStateTest do
     :ok
   end
 
+  test "query a non-existing aggregate_state" do
+    aggregate_uuid = UUID.uuid4()
+
+    assert match?(
+             {:noproc, {GenServer, :call, _request}},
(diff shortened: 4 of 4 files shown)


jvantuyl:
I agree.  It's likely to be more performant, too (if only slightly).

That said, I think this wouldn't be backwards-incompatible.  Anyone that previously detects a missing aggregate correctly will have their code immediately start seeing `nil` values that it never saw before. So maybe it's only appropriate for a major release and should be well documented.  Or, alternatively, it could maybe be feature-flagged with a compile-time config option.

yordis:
The intent here matters a lot, and what do you mean by "the aggregate does not exist"

fahchen:
> The intent here matters a lot, and what do you mean by "the aggregate does not exist"

If an aggregate cannot be reconstructed from events, can we assume that it does not exist?

jvantuyl:
I understand it to mean:  "Asking for an aggregate with an ID that doesn't correspond to any known aggregate."

The intuition here is that returning effectively an empty struct makes it appear that there's an aggregate where there is not really one.  Returning `nil` for `aggregate_state` and `0` for `aggregate_version` makes a lot of sense.

The alternative is that detecting a bad ID is kind of non-intuitive and, if you don't check carefully, you may also start using what is effectively an uninitialized aggregate unknowingly.

fahchen:
> Because if you use nil, you will need to read source code files in production/environment to know the value of status. The worst situation is if two deployments have different values because the business requirements changed and the status value is different.


We can construct the initial state via `stuct(__MODULE__)` if necessary, this is how the aggregate builder does now. https://github.com/commanded/commanded/blob/e20fddd01f46a6668ce7360761c7136e5e3bc497/lib/commanded/aggregates/aggregate_state_builder.ex#L32

You said maybe we need a function to construct the initial state. I think it depends. For now the initial state is determined at compile time, after adding a construct function, the initial state may be determined at runtime. I don’t think it’s a good idea that the construct function generates different initial states at, because this may make the initial state undetermined, and may impact the eventual consistency, may make commands handlers more complicated.

yordis:
I shared my concern about these changes and I am strongly against them, to begin with, it would be a **breaking change** since such `public` function is leverage for different use cases, my personal situation, a Phoenix Dashboard for Commanded and production debugging tooling.

---

> We can construct the initial state via stuct(__MODULE__) if necessary, this is how the aggregate builder does now.

That is exactly what I do not want to be doing in production!

Such technicality should be abstracted away, reasons why some peeps (like me) use `aggregate_state` function, it is not `just` about that!

We are going in a circle by now, I asked you to share your concrete use case rather than bikeshed over code for the sake of code, everything makes sense out of context. I am not saying you are wrong completely, just that these changes are missing out on a lot of contexts and making assumptions.

In the end, It Depends ™️ so I don't think it will be a productive conversation.

yordis:
What do you intend to do in your business?

Like, remove the technicality of the situation for a moment. That is what I am trying to get out of you to comprehend and help you better altogether.

I understand your problem with the existing implementation and the `what` and `how` you are trying to change, but I still don't know the `why` behind it because it could be an XY problem situation here. Please read https://xyproblem.info/ if you need to become more familiar with it.

Why do you need to determine whether the aggregate state is the initial state? ESPECIALLY if you are not building dev tools and it is a domain need (concerning if so), more often than not, people think of the problem in a sub-optimal way.

fahchen:
Here is my situation:
My service could handle requests from outside, from users or other systems, which are handled by aggregate_state function. I have to tell them the aggregate’s existence accurately, for they don’t have context.

FYI: Why this feature introduced, and there is a case that shares with mine, see https://github.com/commanded/commanded/issues/336

jvantuyl:
@fahchen For what it's worth, this is why I tend to have my enums default to an `UNKNOWN` value.  It helps distinguish the zero value from a fresh but initialized value.  I picked this up in Go, but it works for this here.  Not sure if that works for your use case or not, but it's not a terrible approach from my experience.

jvantuyl:
@yordis Maybe I'm just misunderstanding the use-case.  How exactly is the `aggregate_state/3` meant to be used?

jvantuyl:
> @jvantuyl I agree with you that we can have an unknown status as the initial flag. However, if we could identify the initial state, even without altering any bytes of our business code, that would be a just enough approach.

I would support this.  Semantically, `nil` is well understood to mean "doesn't exist".  Making the user craft their own version of this for every aggregate doesn't sit well with me either.  But... yeah... I suggest the workaround mostly for when other people stumble across this and need a solution.

drteeth:
> @yordis Maybe I'm just misunderstanding the use-case. How exactly is the `aggregate_state/3` meant to be used?

I think a lot of the discussion so far is a little bit unclear because of the focus on aggregate_state/3. I think @yordis (and me while I was first reading) missed that we were even talking about aggregate_state/3. There is a focus here on aggregate_state/3 and I think @yordis is right to question what you are trying to do and why aggregate_state/3 seems to be the preferred tool.

To me, it's not something I would reach for unless I had run out of options elsewhere. See @jwilger's comment in #336. You've described your use case and I'm not clear why an event handler/read model wouldn't suit your needs. It would be more idiomatic IMO and you could customize to your heart's content. Most of all you'd be crafting a domain specific solution with existing mechanics instead of trying to bend the mechanics of aggregate_state/3 to your domain problem. Are you sure you can't handle this with a read model? If you elaborate more on your use case, I'm sure someone here (or in the commanded slack channel) can help find you such a solution.

As for the breaking change, I don't see how aggregate_state/3 changing it's signature would be a good thing. A new function gets around that but makes the choice of when to use one over the other unclear.

Finally, while nil/null is widely used in the industry to indicate the absence of a value, it is considered a anti-pattern by many. I would rather see an ok-error tuple if we have to have a change here.

For fun, see null's "creator" regretting its introduction in 1965:
[The Billion Dollar Mistake](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/)

jwilger:
As others have noted, this is a breaking change that doesn't really seem to unlock any additional functionality. Feel free to reopen/restart discussion if you can provide further evidence of something that is impossible and/or extremely unergonomic to implement with the current design.

PR #506, opened 2022-10-05
feat: bubble up the error from aggregate failure

yordis:
## Context

We are getting many of these errors, but we do not have many insights into what exactly happens.
This helps us to assess what is failing.

### DIFF
--- a/lib/commanded/commands/dispatcher.ex
+++ b/lib/commanded/commands/dispatcher.ex
@@ -103,8 +103,8 @@ defmodule Commanded.Commands.Dispatcher do
         {:exit, {{:nodedown, _node_name}, {GenServer, :call, _}}} ->
           {:error, :remote_node_down}
 
-        {:exit, _reason} ->
-          {:error, :aggregate_execution_failed}
+        {:exit, reason} ->
+          {:error, {:aggregate_execution_failed, reason}}
 
         nil ->
           {:error, :aggregate_execution_timeout}


lib/commanded/commands/dispatcher.ex
```diff
@@ -103,8 +103,8 @@ defmodule Commanded.Commands.Dispatcher do
         {:exit, {{:nodedown, _node_name}, {GenServer, :call, _}}} ->
           {:error, :remote_node_down}
 
-        {:exit, _reason} ->
-          {:error, :aggregate_execution_failed}
+        {:exit, reason} ->
+          {:error, {:aggregate_execution_failed, reason}}
```
fahchen, line 107:
```suggestion
          {:error, :aggregate_execution_failed, reason}
```

Returning three elements tuple could be more compatible.


1. the result handles in the next codes can handle these results;
https://github.com/commanded/commanded/blob/7be1943acdbd313707743bcc9142a888ef869dae/lib/commanded/commands/dispatcher.ex#L136-L146
2. by adding new middleware to returning `{:error, {:aggregate_execution_failed, reason}}`, you can get the reason of a `aggregate_execution_failed` error from the dispatching result.

(note: *I'm not a maintainer, just commenting*)

jvantuyl: I've always found "three-element tuples" to be more trouble.  Often, consuming code wants to bubble up errors.  Two-element tuples make that easy to detect.  Three-element tuples require the surrounding code to add possibly many different matches.  As such, I vastly prefer to have the error "reason" become a tuple (or struct) rather than to change the shape of the return value.

slashdotdash:
We really need to a test to demonstrate what this change is doing, verify it works as expected, and to ensure that it cannot be inadvertently broken with any future changes.

drteeth:
@yordis do you still need this one or can it be closed?

yordis:
One of those that you do not need it until you do again! I do not need it at the moment, and hopefully ever again, but I am sure it will be a matter of time

yordis:
I somewhat agree, I do not have any strong opinion, but, in general, it is probably a good idea to bubble up this error; feel free to close it

lib/commanded/commands/dispatcher.ex
jwilger: Furthermore, assuming I'm reading this correctly, with a three-element tuple, the dispatcher would need to be updated to still respond with the actual reason if you want it in your calling code; otherwise the Pipeline response is dropping the reason and only returning the two-element tuple.

jwilger:
Closing as per discussion.

PR #507, opened 2022-10-18
Record aggregate state while processing Multi

joaobalsini:
Hello!

On our company we use commanded a lot, and, while working in a complex multi, I found the need to extend actual functionality to mimic what Ecto allows us to do inside Multis. On one of the steps of the multi, I need to calculate the aggregate state diff between previous step and actual step to know what exactly happened there. Our aggregate processes a lot of information and some information will only be known after the event is processed.

Before this PR, a user could only compare aggregate state before the multi started with the aggregate state on the current step.This PR allows a user to record aggregate state under a name for each of the multi steps (recorded after the step has been processed in the aggregate) and later compare it with the aggregate state before the multi started or after any of the previous steps.

This feature helps on some complex cases multis, in which we need to know what changes were applied on previous steps of the multi. 

Existing multis can be kept as is, just adding more options to it.

### Example
```
      aggregate
      |> Multi.new()
      |> Multi.execute(:first_step, fn _aggregate ->
        %Event{data: 1}
      end)
      |> Multi.execute(fn _aggregate, %{first_step: _aggregate_after_first_step} ->
        %Event{data: 2}
      end)
      |> Multi.reduce(:another_step, [3, 4, 5], fn _aggregate, index, %{first_step: _aggregate_after_first_step} ->
        %Event{data: index}
      end)
      |> Multi.reduce(:last_step, [6, 7], fn _aggregate, index ->
        %Event{data: index}
      end)
```

### DIFF
--- a/lib/commanded/aggregates/multi.ex
+++ b/lib/commanded/aggregates/multi.ex
@@ -158,55 +158,57 @@ defmodule Commanded.Aggregate.Multi do
   aggregate state, the aggregate state for each named step and all created events.
   """
   @spec run(Multi.t()) ::
-          {aggregate :: struct(), steps :: %{atom() => struct()}, list(event :: struct())}
-          | {:error, reason :: any()}
+          {aggregate :: struct(), list(event :: struct())} | {:error, reason :: any()}
   def run(%Multi{aggregate: aggregate, executions: executions}) do
     try do
-      executions
-      |> Enum.reverse()
-      |> Enum.reduce({aggregate, %{}, []}, fn
-        {step_name, execute_fun}, {aggregate, steps, events}
-        when is_function(execute_fun, 1) or is_function(execute_fun, 2) ->
-          case execute_function(execute_fun, aggregate, steps) do
-            {:error, _reason} = error ->
-              throw(error)
+      {evolved_aggregate, _steps, pending_events} =
+        executions
+        |> Enum.reverse()
+        |> Enum.reduce({aggregate, %{}, []}, fn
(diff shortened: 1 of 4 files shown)


lib/commanded/aggregates/multi.ex
```diff
@@ -88,57 +114,122 @@ defmodule Commanded.Aggregate.Multi do
...
 
   @doc """
   Run the execute functions contained within the multi, returning the updated
-  aggregate state and all created events.
+  aggregate state, the aggregate state for each named step and all created events.
   """
   @spec run(Multi.t()) ::
-          {aggregate :: struct(), list(event :: struct())} | {:error, reason :: any()}
+          {aggregate :: struct(), steps :: %{atom() => struct()}, list(event :: struct())}
+          | {:error, reason :: any()}
```
slashdotdash, line 162: Removing the `steps` from the return tuple will allow the function contract to be unchanged and hence won't be a breaking change.

```elixir 
@spec run(Multi.t()) :: {aggregate :: struct(), list(event :: struct())} | {:error, reason :: any()}
```

It should be acceptable to do this since the returned `steps` value isn't actually being used outside of the tests.

joaobalsini: Makes sense, thanks for the feedback! I will prepare a new commit.

joaobalsini: Updates here: 28b089e8f580fdbfe81da2981bbcbbd2f6936620

slashdotdash (approved):

slashdotdash:
Thanks for the PR @joaobalsini.

PR #508, opened 2022-10-24
Reduce compile-time dependencies

mattnenterprise:
Moves opts out of module attributes for process managers and event handlers. The opts can reference modules which will cause [compile-time dependencies](https://hexdocs.pm/elixir/1.14/Kernel.html#@/1-attention-compile-time-dependencies) if they are in module attributes.


slashdotdash:
Thanks for the pull request Matt.

PR #511, opened 2022-11-16
Remove compile dependencies in Router

dvic:
- Remove compile time deps for Router.dispatch/2
- Remove compile time deps for Router.identify/2
- Remove compile time deps for Router.middleware/1
- Optimize identities lookup

### DIFF
--- a/lib/commanded/commands/router.ex
+++ b/lib/commanded/commands/router.ex
@@ -591,9 +591,13 @@ defmodule Commanded.Commands.Router do
         @middleware
       end
 
+      @registered_identities Map.new(
+                               @registered_identity_refs,
+                               fn {ref, opts} -> {Module.concat(ref), opts} end
+                             )
+
       defp __get_registered_identity__(aggregate_module) do
-        aggregate_module_ref = Module.split(aggregate_module)
-        Map.get(@registered_identity_refs, aggregate_module_ref)
+        Map.get(@registered_identities, aggregate_module)
       end
 
       # Catch unregistered commands, log and return an error.
--- a/lib/commanded/commands/router.ex
+++ b/lib/commanded/commands/router.ex
@@ -220,7 +220,7 @@ defmodule Commanded.Commands.Router do
 
       Module.register_attribute(__MODULE__, :registered_command_refs, accumulate: true)
       Module.register_attribute(__MODULE__, :registered_middleware, accumulate: true)
-      Module.register_attribute(__MODULE__, :registered_identities, accumulate: false)
+      Module.register_attribute(__MODULE__, :registered_identity_refs, accumulate: false)
 
       @default_dispatch_opts [
         application: Keyword.get(unquote(opts), :application),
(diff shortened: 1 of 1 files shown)


lib/commanded/commands/router.ex
```diff
@@ -263,8 +277,11 @@ defmodule Commanded.Commands.Router do
 
   """
   defmacro middleware(middleware_module) do
+    middleware_module = :elixir_aliases.expand_or_concat(middleware_module, __CALLER__)
```
slashdotdash, line 280: Where does the `:elixir_aliases.expand_or_concat/2` function come from, I can't find it mentioned in the Elixir docs and is it safe to take a dependency on it?

slashdotdash (commented):
Is there a reason why we wouldn't want to recompile a router module if one of its dependencies has been modified, it could have been removed for example and if so we would want the router compilation to fail. 

How can we write a test for this changed behaviour to check that it works as intended?

My preference would be to try to keep the macro code as simple as possible since it is already difficult enough to understand, verify it works as expected, and to support new features in the future more easily.

lib/commanded/commands/router.ex
dvic: It's used in Macro.expand, see https://github.com/elixir-lang/elixir/blob/v1.14.2/lib/elixir/lib/macro.ex#L1771

I guess it's a private api 😅 but I did not find another way to get the full module name without introducing a compile-time dependency on it.

dvic:
> Is there a reason why we wouldn't want to recompile a router module if one of its dependencies has been modified, it could have been removed for example and if so we would want the router compilation to fail.

Because the router module does not depend on the contents of the aggregate module or command modules, it just basically needs the name to generate a function clause and check for duplicates. If a module is removed, then the compilation will fail because the generated code references a non-existing struct.

> How can we write a test for this changed behaviour to check that it works as intended?

I'm not sure how to add a test for this but I verified it manually and I got a compile error after removing a command that previously successfully compiled.

> My preference would be to try to keep the macro code as simple as possible since it is already difficult enough to understand, verify it works as expected, and to support new features in the future more easily.

I agree that the code has become more complex, but I find it a decent trade-off for the benefit. On our "smallish" project with 16 commands and 2 aggregates we went from 74 to 52 compile-connected dependencies. But I understand your doubt.

I have to add that the biggest win did come from the fact that we switched form dispatching on the application (which had two routers) to dispatching to the routers directly, thereby removing additional compile dependencies introduced by CompositeRouter (which can't be removed with the current implementation).

Do you have any experience with slow compile times on large projects? For example, in our case change a command would trigger 34 files to recompile. It really slowed down the editor feedback loop.

lib/commanded/commands/router.ex
slashdotdash: I don't think we want to take a dependency on an internal (private) function.

slashdotdash:
> Do you have any experience with slow compile times on large projects?

No, it's not something I can say has ever caused me an issue. I typically create a router module per "context" within an application so they each only have a few commands and aggregates registered. 

I am inclined to not merge this PR due to the added complexity for a small improvement which can be resolved by using composite routers which is a better design (IMO).

dvic:
> I am inclined to not merge this PR due to the added complexity for a small improvement which can be resolved by using composite routers which is a better design (IMO). The additional complexity - which is difficult to understand - is something that I will be forced to maintain indefinitely!

I agree 👍

PR #512, opened 2022-11-23
Properly handle EXIT signal in event handler

dvic:


dvic:
This solves an issue we faced with the `:global` registry (i.e., on deploy we would have many errors reported in sentry).

yordis:
Oh yes! This reporting! Not all 🦸🏻‍♂️ heroes wear capes! Some create PRs!

Maybe add an entry here: https://github.com/commanded/commanded/blob/master/CHANGELOG.md

lib/commanded/event/handler.ex
```diff
@@ -820,6 +820,12 @@ defmodule Commanded.Event.Handler do
     {:stop, reason, state}
   end
 
+  @doc false
+  @impl GenServer
+  def handle_info({:EXIT, _pid, reason}, state) do
+    {:stop, reason, state}
+  end
```
yordis, line 827:
```suggestion
  def handle_info({:EXIT, _pid, reason}, state) do
    Logger.debug(fn -> describe(state) <> " subscription EXIT due to: #{inspect(reason)}" end)
    {:stop, reason, state}
  end
```

yordis (changes requested):

lib/commanded/event/handler.ex
dvic: Good catch :) But it's already logged as part of the `terminate/2` callback

dvic:
> Oh yes! This reporting! Not all 🦸🏻‍♂️ heroes wear capes! Some create PRs!
> 
> Maybe add an entry here: https://github.com/commanded/commanded/blob/master/CHANGELOG.md

hehe thanks, I'm also glad to have found this one, it was driving me nuts :)

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v1.4.2 (unreleased)
```
slashdotdash, line 3: I just use the "Next release" header for unreleased changes as the version number won't be known until preparing the release and depending upon what it contains (breaking changes or minor bugfixes).

slashdotdash (commented):
Can we include a test to demonstrate what this change is doing or fixing and to prevent future regressions.

dvic:
> Can we include a test to demonstrate what this change is doing or fixing and to prevent future regressions.

Done!

yordis (approved):
🚀 🚀 🚀 🚀

yordis:
@slashdotdash any opportunity to merge and release the PR? It will help a lot with the logs at the moments deploying new versions of the system 🙏🏻

drteeth:
For me (not that my opinion counts for anything here) I not clear on what the change is doing, or what's causing the problem. I get that you are seeing unhandled message in your logs, but I don't see how that is tied to shutdowns. Can you provide a bit more backstory and maybe we can see about making the test make more sense. 

What is the problem? Why does it need to be fixed? How does the proposed solution work? What potential consequences does the fix introduce?

Can you make that clear in:
A) The description of the PR with a write-up that describes the problem as it is encountered, and the reason that trapping exits is the best solution. 
B) The test somehow, either with a clear test description, or comments. Maybe like a Given, When, Then?

As it is you are putting the burden of understanding the reasoning for the fix in @slashdotdash's hands. If I were merging this PR, I'd want to know so I can judge what other consequences I might need to be aware of. 

I'm not saying that will get your PR merged, but it might improve your chances.

yordis:
The logs and Sentry are filled with the following.

```
level=error message="MyApp.RdfiAchTransfers.PushTransferLimitUpdater received unexpected message: {:EXIT, #PID<70713.10547.0>, :noconnection}"
level=error message="MyApp.CloseAccountTransfers.TransferPoster received unexpected message: {:EXIT, #PID<70713.19085.0>, :noconnection}"
level=error message="MyApp.CloseAccountTransfers.TraceNumberAssigner received unexpected message: {:EXIT, #PID<70713.19085.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.CustomerConsentVerifier received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.CloseAccountTransfers.FileUploadInitiator received unexpected message: {:EXIT, #PID<70713.19085.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.InitialInterestSetter received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.CloseAccountTransfers.AchGatewayPoster received unexpected message: {:EXIT, #PID<70713.19085.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.IdentificationNumberAssigner received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.CustomerCreator received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.LimitInitializer received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.ProviderAccountCreator received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.AccountFlagRequester received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.InitialDepositAccountNumberAssigner received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.ProviderAccountClosedMarker received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.PendingClosureStatusUpdater received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.ProviderAccountPendinagClosureMarker received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.PendingClosureFlagRequester received unexpected message: {:EXIT, #PID<70713.19127.0>, :noconnection}"
level=error message="MyApp.DepositAccountOpenings.LimitInitializer received unexpected message: {:EXIT, #PID<70713.19127.0>, :shutdown}"
level=error message="MyApp.DepositAccountOpenings.IdentificationNumberAssigner received unexpected message: {:EXIT, #PID<70713.19127.0>, :shutdown}"
level=error message="MyApp.CloseAccountTransfers.FileUploadInitiator received unexpected message: {:EXIT, #PID<70713.19085.0>, :shutdown}"
level=error message="MyApp.DepositAccountOpenings.InitialInterestSetter received unexpected message: {:EXIT, #PID<70713.19127.0>, :shutdown}"
level=error message="MyApp.DepositAccountOpenings.CustomerConsentVerifier received unexpected message: {:EXIT, #PID<70713.19127.0>, :shutdown}"
level=error message="MyApp.DepositAccountOpenings.ProviderAccountCreator received unexpected message: {:EXIT, #PID<70713.19127.0>, :shutdown}"
level=error message="MyApp.DepositAccountOpenings.CustomerCreator received unexpected message: {:EXIT, #PID<70713.19127.0>, :shutdown}"
```

Every deployment will cause those errors (which that is not even the full list of Event Handlers)

dvic:
> For me (not that my opinion counts for anything here) I not clear on what the change is doing, or what's causing the problem. I get that you are seeing unhandled message in your logs, but I don't see how that is tied to shutdowns. Can you provide a bit more backstory and maybe we can see about making the test make more sense.
> 
> What is the problem? Why does it need to be fixed? How does the proposed solution work? What potential consequences does the fix introduce?
> 
> Can you make that clear in: A) The description of the PR with a write-up that describes the problem as it is encountered, and the reason that trapping exits is the best solution. B) The test somehow, either with a clear test description, or comments. Maybe like a Given, When, Then?
> 
> As it is you are putting the burden of understanding the reasoning for the fix in @slashdotdash's hands. If I were merging this PR, I'd want to know so I can judge what other consequences I might need to be aware of.
> 
> I'm not saying that will get your PR merged, but it might improve your chances.

The reason for this PR is that in the `init/1` function the exits are being trapped call to `Process.flag(:trap_exit, true)`. This means that shutdown signals coming from a process other than the process that started this process, will result in a `{:EXIT, _pid, reason}` message being sent (shutdown from the parent process will result in the `terminate` callback being called). 

I have not investigated why the exits are being trapped or what process is sending the exit signals, but it is definitely happening as @yordis can also confirm. Hope this helps clear up the intent of the PR.

lib/commanded/event/handler.ex
```diff
@@ -820,6 +820,17 @@ defmodule Commanded.Event.Handler do
...
+  @doc false
+  @impl GenServer
+  def handle_info({:EXIT, _pid, reason}, state) do
+    if reason == :normal do
+      # linked process exited normally, don't shutdown
+      {:noreply, state}
+    else
+      {:stop, reason, state}
+    end
+  end
```
yordis, line 832:
```suggestion
  @doc false
  @impl GenServer
  def handle_info({:EXIT, _pid, :normal}, state) do
    # linked process exited normally, don't shutdown
    {:noreply, state}
  end
  
  @impl GenServer
  def handle_info({:EXIT, _pid, reason}, state) do
    {:stop, reason, state}
  end
```

yordis (changes requested):

lib/commanded/event/handler.ex
```diff
@@ -821,14 +821,15 @@ defmodule Commanded.Event.Handler do
   end
 
   @doc false
+  @impl GenServer
+  def handle_info({:EXIT, _pid, :normal}, state) do
+    # linked process exited normally, don't shutdown
+    {:noreply, state}
+  end
```
yordis, line 828: What would be the implications when it is gracefully shutting down?

You mention "don't shutdown," but if it is gracefully shutting down, I assume you should stop the process, no? How that work here?

dvic: If it's gracefully shutting down, meaning being shut down by it's parent (supervisor) process, then the `terminate` callback is called, and no signal is sent. These signals are only for linked processes other than the parent, and for those we don't want the handler to shutdown when they exit.

The background story of this: https://github.com/bitcrowd/chromic_pdf/issues/233, in short: I was generating a pdf inside an event handler and in tests it runs in on demand mode, which means it starts and stops a supervisor/genserver just for that one request for generating a pdf. This shut down the event handler and made the tests fail in some situations.

yordis:
@dvic, the CI is failing due to a formatting issue.

@slashdotdash is there any opportunity to support @dvic in the feature? Can we do anything to support you in maintaining the package?

lib/commanded/event/handler.ex
tomekowal: I believe this is a good change :+1: 

Background:
By default, the process is not trapping exits and then linking works like this for a link between A and B:
a) If process A dies with reason `:normal`, it means it finished itself, and everything is fine, so BEAM does NOT kill B.
b) If process A dies with any other reason, BEAM kills process B with the same reason.

However, `Commanded.Event.Handler` needs to trap exits to clean up in the `terminate` callback.
Beam then translates exit signals into messages.
1. If the message comes from the parent (the supervisor), the `terminate` callback is called.
2. Otherwise, the message comes from another linked process (e.g. someone `spawn_link`ed a process from within the handler.
Without this PR, all those exit signals would be logged as errors in the next `handle_info` clause (line 837) and worse, if a linked process died, the handler would not crash as expected.

So this PR brings back the sane default behaviour from a regular process that does not trap exits:
- the `{:EXIT, _pid, :normal}` covers case a)
- the `{:EXIT, _pid, reason}` covers case b)

tomekowal (approved):

tomekowal:
Hey, @dvic I was just looking through this PR, and it looks like a solution to exactly the same issue we are facing.
I've added a comment in the code further explaining why this is a good change.
Would you consider fixing the formatting so that CI becomes green?

dvic:
> Hey, @dvic I was just looking through this PR, and it looks like a solution to exactly the same issue we are facing.
> I've added a comment in the code further explaining why this is a good change.
> Would you consider fixing the formatting so that CI becomes green?

Done!

predrag-rakic:
I'd also like this PR merged.
Thank you.

yordis:
@dvic, if it is not much to ask, maybe add a new entry to https://github.com/commanded/commanded/blob/master/CHANGELOG.md. It would help to keep track of what will be part of the release.

I think 😄

Issue #440, opened 2021-04-29
Allow access to event metadata in process manager callback functions


slashdotdash:
Extend the existing process manager `interested?/1` and `handle/2` callback functions to allow access to metadata by optionally supporting a `interested?/2` or `handle/3`. These new functions would also passed the event metadata as the last function parameter. 

These changes would be backwards compatible so that either a `interested?/1` or `interested?/2` and `handle/2` or `handle/3` function could be implemented for each event.

### Example usage

```elixir
defmodule ExampleProcessManager do
  use Commanded.ProcessManagers.ProcessManager,
    application: ExampleApp,
    name: "ExampleProcessManager"

  def interested?(%AnEvent{uuid: uuid}), do: {:start, uuid}
  def interested?(%AnotherEvent{uuid: uuid}), do: {:continue, uuid}

  def handle(%ExampleProcessManager{}, %AnEvent{}, metadata) do
    # `handle/3` function when metadata is needed
  end

  def handle(%ExampleProcessManager{}, %AnotherEvent{}) do
    # `handle/2` function when metadata is not required
  end
end
```

vasspilka:
I could possibly take a look into this, by the way maybe we can close #327 in favor of this

bitwalker:
I could use this in something I'm working on myself, I'm happy to lend a hand as well if there is a preferred approach.

dvic:
@slashdotdash any chance that #514 will make it soon? I tested it also locally and it LGTM

PJKerrigan:
This would be very useful.

It's often necessary to know when an event was created, and that's stored in the metadata. Adding a separate field for this on an event means that you have two dates, often with slightly different values, meaning the same thing.

Having the metadata when an event is processed or applied would be wonderful.

slashdotdash:
Done in #514.

PR #514, opened 2022-11-30
Process manager metadata read access

vheathen:
The PR addresses #440 and allows metadata read access from all callbacks except `error/3` (as the last one already has access to full RecordedEvent struct).

Supports both old and new callback versions (with and without metadata as the last parameter)

Also adds copying metadata from processed events to new commands by default. 

Modification of the metadata inside the process manager `handle/3` and `handle/2` callbacks are subject of the possible next PR.

### DIFF
--- a/lib/commanded/event_store/recorded_event.ex
+++ b/lib/commanded/event_store/recorded_event.ex
@@ -34,18 +34,55 @@ defmodule Commanded.EventStore.RecordedEvent do
 
   @type uuid :: String.t()
 
+  @type domain_event :: struct()
+
+  @type event_id :: uuid()
+  @type event_number :: non_neg_integer()
+  @type stream_id :: String.t()
+  @type stream_version :: non_neg_integer()
+  @type causation_id :: uuid() | nil
+  @type correlation_id :: uuid() | nil
+  @type event_type :: String.t()
+  @type data :: domain_event()
+  @type metadata :: map()
+  @type created_at :: DateTime.t()
+
   @type t :: %RecordedEvent{
-          event_id: uuid(),
-          event_number: non_neg_integer(),
-          stream_id: String.t(),
-          stream_version: non_neg_integer(),
-          causation_id: uuid() | nil,
-          correlation_id: uuid() | nil,
-          event_type: String.t(),
-          data: struct(),
-          metadata: map(),
-          created_at: DateTime.t()
+          event_id: event_id(),
+          event_number: event_number(),
+          stream_id: stream_id(),
+          stream_version: stream_version(),
+          causation_id: causation_id(),
+          correlation_id: correlation_id(),
+          event_type: event_type(),
+          data: data(),
+          metadata: metadata(),
+          created_at: created_at()
         }
+
+  @type enriched_metadata :: %{
+          :event_id => event_id(),
+          :event_number => event_number(),
+          :stream_id => stream_id(),
+          :stream_version => stream_version(),
+          :correlation_id => correlation_id(),
+          :causation_id => causation_id(),
+          :created_at => created_at(),
+          optional(atom()) => term(),
+          optional(String.t()) => term()
+        }
+
+  @type additional_fields_to_deplete :: [atom()]
+
+  @enriched_metadata_fields ~w(
+    event_id
+    event_number
+    stream_id
+    stream_version
+    correlation_id
+    causation_id
+    created_at
+  )a
 
   defstruct [
     :event_id,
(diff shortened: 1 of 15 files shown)


jwilger (approved):
@slashdotdash Any objections to getting this merged in? I actually came here today thinking about making the same (basically) change, because I also need to have metadata passed along in order to support authorization middleware used in the command dispatch.

dvic:
The only thing I might want to be optional:

> Also adds copying metadata from processed events to new commands by default.

jwilger:
> The only thing I might want to be optional:
> 
> > Also adds copying metadata from processed events to new commands by default.

Good call. Perhaps when using handle/3 we could return commands as something like:

```
[some_command, {some_other_command, modified_metadata}, ...]
```

where the first one would just get the metadata copied over (as this PR does now), and the second command would allow you to modify that metadata first.

jwilger:
@vheathen: what do you think about that suggestion? I realize it's been quite a while since you submitted the PR, would you want to make that change? If not, I'd be happy to pick it up and give it a go.

vheathen:
@jwilger I was going to open a discussion about modifying metadata during the second iteration after this PR will be accepted (or rejected). I also thought to return a commands list exactly in the format you suggested above but just wanted to debate it with the community first: may be someone would offer a better approach. But I can add this functionality to the current PR if its ok.

@dvic About copying metadata to the issued commands by default: I was thinking that it is the most logical at least until it will be possible to modify metadata. May be I'm a little bit biased here, but I think that metadata is usually used to keep information which is not a part of the event (because it is outside of the domain itself) but is necessary to track commands-events chain (when relation/causation id are not enough) or for some technical reasons. As an example: we keep a tenant id in metadata, and the only reason we need it is technical, the subdomain knows nothing about tenants, and there is no sense to add a tenant id to every event and command. WDYT?

dvic:
> @dvic About copying metadata to the issued commands by default: I was thinking that it is the most logical at least until it will be possible to modify metadata. May be I'm a little bit biased here, but I think that metadata is usually used to keep information which is not a part of the event (because it is outside of the domain itself) but is necessary to track commands-events chain (when relation/causation id are not enough) or for some technical reasons. As an example: we keep a tenant id in metadata, and the only reason we need it is technical, the subdomain knows nothing about tenants, and there is no sense to add a tenant id to every event and command. WDYT?

I can see the reasons for wanting to have the metadata passed along, I'm just not sure if it should be enabled by default. In our case, we solve the multi-tenancy by having separate event stores and commanded instances, but we definitely have use-cases for wanting to pass along the metadata or modifying it.

In any case, what about having an optional callback that determines the dispatched metadata (they `metadata` keyword of the dispatch function)?

```elixir
@callback dispatch_metadata(received_event :: event(), dispatched_commands :: [command()]) :: map()
```

slashdotdash:
This change adds complexity to a the process manager with the overloading of the various callback functions. 

I would say that any information needed by a process manager (or aggregate) to make a decision should be included in the event payload, not metadata. Doing so makes this PR unnecessary as the relevant data would be accessible from the event. It is also simpler to have to only deal with events, not events and their associated metadata.

vheathen:
@slashdotdash 

I would kindly ask to explain why do we have metadata at all then? Following that logic, we should put everything to the events themselves. 

I see metadata as a container for technical/supplementary data which don't have any sense in the domain but can be necessary for technical implementation or other needs. As an example, I will repeat from above: our events metadata got a tenant id, a timestamp of the moment when command was dispatched and an original user id. In many cases these fields don't have any sense in the domain as it knows nothing about tenants, users (because domain events have their actors inside - and only when they are necessary in terms of domain), timestamps, etc, and the reasons we need it: 

- a tenant id is to partition read model;
- a user id for ACL; 
- all of these fields including timestamp are necessary for debugging and auditing.

At any point in time, it is possible to get rid of the metadata and domain events will still make sense from the domain perspective. Two main places where metadata is used: in the projectors and in the ACL middleware. At the same time, we don't contaminate events with irrelevant _for the domain_ information.

So now we have a situation where manually dispatched commands are supplied with required metadata, directly resulting events have that metadata but subsequent commands after those events (and their events) from the process manager don't. Why?

Another thing is that process managers might need to modify that metadata - for example, to issue a command as a system user and not as an original user (or as nobody if it is not possible to copy that data from the original event). This PR didn't include any modification, only copying, but that can be useful as well and I had an intention to implement it after decision on the current one.

Also, the situations possible where it is necessary to keep state in the database because it's too large or complex or it is impossible to address things from the event itself - I don't like this approach, but already had a case where it was necessary to do this way at least during transition period (because of the legacy bad event design, but it is what it is). And - again, because of the technical reasons - metadata can be necessary to correctly access db.

I totally agree about aggregates: these are exclusively domain creatures, but process managers aren't, at least if we have an idea of metadata at all. Even when we speak about aggregates: they are copying metadata from commands to events. So why process managers (which are the opposite of aggregate in some sense) don't? 
Though the main difference between aggregates and PMs in terms of metadata is that metadata built before dispatching a command, so an aggregate don't need worry about it. The process manager dispatches commands though, so it also need to be able to build metadata. And to do so it needs old metadata access.

What do you think?

PR #522, opened 2023-01-25
Add docs for event handler modules consistency dispatch option

MrYawe:
I noticed that there is an inconsistency (😄) between the [wiki](https://github.com/commanded/commanded/blob/master/guides/Commands.md#command-dispatch-consistency-guarantee) and the commanded documentation.

Feel free to change the wording.


slashdotdash:
Thanks for the pull request Yannis.

PR #524, opened 2023-02-03
Adds devcontainer configuraton for development environments

jwilger:
This makes it trivial to get a development environment set up using either VSCode or GitHub Codespaces (or any other editor/IDE that supports the devcontainer standards).

(N.B. this is for developers working on commanded itself vs. developers working on an application that uses commanded.)


slashdotdash:
This looks like it could be a pain for someone (likely me!) to maintain going forwards. I'd rather not have to support a Dockerfile and docker-compose just for development, unless there is a strong motivation for doing so. Can you provide more context as to what benefits this change brings?

PR #528, opened 2023-02-21
Support opts in append_to_stream

dvic:


lib/commanded/event_store/adapter.ex
```diff
@@ -31,7 +31,8 @@ defmodule Commanded.EventStore.Adapter do
               adapter_meta,
               stream_uuid,
               expected_version,
-              events :: list(EventData.t())
+              events :: list(EventData.t()),
+              options
```
janpieper, line 35: Wouldn't this mean that we would get a warning for every custom adapter that does not implement `append_to_stream/5`? And additionally warnings for `append_to_stream/4` functions using `@impl true` :thinking:

PR #529, opened 2023-03-03
Allow setting supervising options such as `max_restarts` for aggregate supervisor.

qhwa:
Hi, this is a PR to support custom supervisor options for aggregates.

## Context

Our application has many aggregates, each of which listens to a stream, so we want to set a higher `max_restarts` option for the dynamic supervisor `Commanded.Aggregates.Supervisor`, so that a few failures won't crash the whole supervision tree.

This patch allows applications to specify custom supervising options, as the following example shows:

```elixir
config :my_app, MyApp.EventApplication,
  aggregate_supervisor_options: [
    max_restarts: 1_000,
    max_seconds: 10
  ]
```

Your feedback is welcome!


qhwa:
There are some pending things regarding this change:

* [ ] documentation
* [ ] testing

But we can talk briefly if this is a good idea or if there is a better approach. What do you think?
@thetamind @jwilger @patrickdet

PR #534, opened 2023-06-02
add compile-time config option to globally set aggregate lifespan

jvantuyl:
The default behavior of "aggregate processes stay around forever" can present difficulty in a production environment.  For any system that creates a steady stream of new aggregates, memory creeps up slowly until exhausted.

It is currently possible to override this on a per-router basis, but that only creates a situation where forgetting to specify it one place creates an even more subtle "memory leak".

This commit allows changing the default globally at compile-time.  It does not allow changing it at run-time, as this seemed like it would cause unpredictable behavior and possibly negatively impact performance.

Testing this is difficult since it's a compile-time change.  As a compromise, I set the value to the default explicitly in the testing config.  While this doesn't fully test the functionality, it does at least work-out the code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+- Add configuration option `:aggregate_default_lifespan` to change the default lifespan module globally at compile-time (link-forthcoming).
+
 ## v1.4.2
 
 -  Record aggregate state while processing `Commanded.Aggregate.Multi` ([#507](https://github.com/commanded/commanded/pull/507)).
--- a/config/test.exs
+++ b/config/test.exs
@@ -31,3 +31,5 @@ config :commanded, Commanded.ProcessManagers.ExampleApp, default_app_config
 config :commanded, Commanded.ProcessManagers.ResumeApp, default_app_config
 config :commanded, Commanded.ProcessManagers.TodoApp, default_app_config
 config :commanded, Commanded.TestApplication, default_app_config
+
+config :commanded, :aggregate_default_lifespan, Commanded.Aggregates.DefaultLifespan
--- a/guides/Commands.md
+++ b/guides/Commands.md
@@ -415,6 +415,8 @@ Return `:stop` or `{:stop, reason}` to immediately shutdown the aggregate proces
 
 You can also return `:hibernate` and the process is hibernated, it will continue its loop once a message is in its message queue. Hibernating an aggregate causes garbage collection and minimises the memory used by the process. Hibernating should not be used aggressively as too much time could be spent garbage collecting.
 
+If you wish to change the default, set the `:aggregate_default_lifespan` option for the `:commanded` application.  This must be done at compile-time.  It cannot be changed at runtime.
+
 ## Middleware
 
 Allows a command router to define middleware modules that are executed before and after success or failure of each command dispatch.
--- a/lib/commanded/aggregates/execution_context.ex
+++ b/lib/commanded/aggregates/execution_context.ex
@@ -33,7 +33,9 @@ defmodule Commanded.Aggregates.ExecutionContext do
     - `lifespan` - a module implementing the `Commanded.Aggregates.AggregateLifespan`
       behaviour to control the aggregate instance process lifespan. The default
       value, `Commanded.Aggregates.DefaultLifespan`, keeps the process running
-      indefinitely.
+      indefinitely.  The default can be overridden by setting the
+      `:aggregate_default_lifespan` configuration for the `:commanded` application
+      at compile-time.  Cannot be reconfigured at run-time.
 
   """
 
@@ -41,6 +43,12 @@ defmodule Commanded.Aggregates.ExecutionContext do
   alias Commanded.Aggregates.DefaultLifespan
   alias Commanded.Aggregates.ExecutionContext
   alias Commanded.Commands.ExecutionResult
+
+  @default_lifespan Application.compile_env(
+                      :commanded,
+                      :aggregate_default_lifespan,
+                      DefaultLifespan
+                    )
 
   defstruct [
     :command,
@@ -51,7 +59,7 @@ defmodule Commanded.Aggregates.ExecutionContext do
     before_execute: nil,
     retry_attempts: 0,
     returning: false,
-    lifespan: DefaultLifespan,
+    lifespan: @default_lifespan,
     metadata: %{}
   ]
 


yordis (approved):
💜 🚀

jvantuyl:
Any idea what I need to do to get this merged?

fahchen:
I think `Commanded.Application.Config` might be a suitable place to store `aggregate_default_lifespan`, which allows applications to decide whether to set it at runtime or compile-time.

yordis:
These types of configurations are a danger to be swapped at runtime, IMO.

If somebody comes across an actual use case, maybe at that point could be refactored to allow it.

fahchen:
I think that these types of configurations are actually used by commanded applications only.

I created a commit to demo the changes:
https://github.com/fahchen/commanded/commit/fcd0209b6e158d567fb15cc5be98e9dad4ef1db6

yordis:
Hey, @slashdotdash Is there anything we could do to help you to get some love over here and get this feature released?

jvantuyl:
If we can, I'd like to separate the "compile-time vs run-time" discussion from this PR.  The compile-time option is useful enough to include whether or not we later want to add the option to change at run-time.

In general, I'm of the opinion that frameworks should generally have "safe" defaults.  This timeout behavior is effectively the only "non-safe" default I can think of in Commanded.  I wouldn't be opposed to even adding a default timeout (maybe next version), since it shouldn't break most existing code.  At most, it'll have some performance impact for people with very large numbers of events to a single aggregate and no snapshoting.

But, yeah... if we can't have a safe default, it would be nice to at least be able to globally change the default for a given app so nothing can slip through unnoticed.

fahchen:
I think this PR may not be a good approach. 
Firstly, each application has the responsibility for its own configuration, other applications should not depend on that.(official document states at https://hexdocs.pm/elixir/main/Application.html#module-the-application-environment).
Secondly, the config we try to set should belong to a commanded application, not the commanded itself.

The approach I mentioned before it’s not about choosing between compile-time and runtime config, it’s try to respect the two reasons above.

yordis:
I see that you're looking to configure a specific value and you're wondering at which level this could be done. Well, you have a few options, each offering different degrees of control.

1. **Globally:** If you're after a broad solution, you can set it globally like this:
  ```elixir
  config :commanded, :aggregate_default_lifespan, Commanded.Aggregates.DefaultLifespan
  ```
3. **Per Application:** Looking to configure the value on a per-application basis? Here's how you can do that:
  ```elixir
  config :myapp, MyApp.CommandedApp,
    aggregate_default_lifespan: Commanded.Aggregates.DefaultLifespan
  ```
5. **Per Router:** Or, perhaps, you'd prefer to control it on a router level? That can be done in the following way:
  ```elixir
  defmodule MyApp.BankAccount.Router do
    use Commanded.Commands.Router
    
    lifespan BankAccountAggregate, Commanded.Aggregates.DefaultLifespan
  end
  ```
  
So, depending on where you want to exercise control and your needs, you can choose the most suitable option.

---

> Secondly, the config we try to set should belong to a commanded application, not the commanded itself.

You suggested that these should belong to a specific commanded application, rather than commanded itself. I see where you're coming from and agree that having control over such settings per application could indeed be quite useful.

Consider, for instance, configuring it according to the Router. That could be a very viable option as well.

---

> Firstly, each application has the responsibility for its own configuration, other applications should not depend on that.(official document states at https://hexdocs.pm/elixir/main/Application.html#module-the-application-environment).

Let's address the point about each application being responsible for its own configuration. 

You're absolutely right that this is the general guideline outlined in the Elixir documentation. However, it's crucial to remember that this is a guideline and not a rule set in stone.

In reality, there are configurations that are global to `commanded`, such as:

```elixir
config :commanded, type_provider: MyApp.TypeProvider
```

So, the Elixir documentation's statement regarding "application" needs to be understood in the right context. Elixir is referring to OTP applications, since he is indeed adding to `:commanded` application scope, he is respecting and following the guidelines.

It all boils down to the context and specific needs. There's no one-size-fits-all solution here. In other words, "It Depends ™️."

Happy coding! 💜 💜 💜

yordis (changes requested):
@jvantuyl would you be OK trying to also support the configuration at the `Commanded.Application` instance level?

jvantuyl:
For the use-case I'm most concerned with, Application-level would work.  So I'll take it.

Still not a fan of unsafe defaults.  I'd rather not have "Every Commanded Application, with default settings, is a time bomb." be a true statement.  And, given how most applications don't see the load in staging that they do in production... it's a time bomb that's most likely to go off in production.  It's not great.

fahchen:
Maybe we could divide these problems into two parts.
One is to support the default lifespan for aggregates, another is to attempt to fix the time bomb, or alternatively, leave it untouched.

fahchen:
> I see that you're looking to configure a specific value and you're wondering at which level this could be done. Well, you have a few options, each offering different degrees of control.
> 
> 
> 
> 1. **Globally:** If you're after a broad solution, you can set it globally like this:
> 
>   ```elixir
> 
>   config :commanded, :aggregate_default_lifespan, Commanded.Aggregates.DefaultLifespan
> 
>   ```
> 
> 3. **Per Application:** Looking to configure the value on a per-application basis? Here's how you can do that:
> 
>   ```elixir
> 
>   config :myapp, MyApp.CommandedApp,
> 
>     aggregate_default_lifespan: Commanded.Aggregates.DefaultLifespan
> 
>   ```
> 
> 5. **Per Router:** Or, perhaps, you'd prefer to control it on a router level? That can be done in the following way:
> 
>   ```elixir
> 
>   defmodule MyApp.BankAccount.Router do
> 
>     use Commanded.Commands.Router
> 
>     
> 
>     lifespan BankAccountAggregate, Commanded.Aggregates.DefaultLifespan
> 
>   end
> 
>   ```
> 
>   
> 
> So, depending on where you want to exercise control and your needs, you can choose the most suitable option.
> 
> 
> 
> ---
> 
> 
> 
> > Secondly, the config we try to set should belong to a commanded application, not the commanded itself.
> 
> 
> 
> You suggested that these should belong to a specific commanded application, rather than commanded itself. I see where you're coming from and agree that having control over such settings per application could indeed be quite useful.
> 
> 
> 
> Consider, for instance, configuring it according to the Router. That could be a very viable option as well.
> 
> 
> 
> ---
> 
> 
> 
> > Firstly, each application has the responsibility for its own configuration, other applications should not depend on that.(official document states at https://hexdocs.pm/elixir/main/Application.html#module-the-application-environment).
> 
> 
> 
> Let's address the point about each application being responsible for its own configuration. 
> 
> 
> 
> You're absolutely right that this is the general guideline outlined in the Elixir documentation. However, it's crucial to remember that this is a guideline and not a rule set in stone.
> 
> 
> 
> In reality, there are configurations that are global to `commanded`, such as:
> 
> 
> 
> ```elixir
> 
> config :commanded, type_provider: MyApp.TypeProvider
> 
> ```
> 
> 
> 
> So, the Elixir documentation's statement regarding "application" needs to be understood in the right context. Elixir is referring to OTP applications, since he is indeed adding to `:commanded` application scope, he is respecting and following the guidelines.
> 
> 
> 
> It all boils down to the context and specific needs. There's no one-size-fits-all solution here. In other words, "It Depends ™️."
> 
> 
> 
> Happy coding! 💜 💜 💜

IMO, If we can handle this under the guideline, why should we break it?  :)

yordis:
> I'd rather not have "Every Commanded Application, with default settings, is a time bomb."

@jvantuyl, I do not think that is a valid concern at all.

To begin with, I would like to share some history about it based on HMB.

It was critical to have such "sharding" opportunities at multiple levels and understand how scalable the codebase could be as the application performance changes without having to rearchitect or change the system drastically. It was part of the decision-making and reasons to adopt an entirely new stack.

TL;DR because of the following https://10consulting.com/2021/03/18/commanded-application-architecture/

By design, it is good these level configurations.

---

Your concerns are valid, but it is a programmer problem rather than a technology problem. Commanded already allows you to do multi-app architecture styles.

yordis:
> IMO, If we can handle this under the guideline, why should we break it? :)

@fahchen I do not think you understand honestly, but it is not that important. Yes, I agree with you, your proposal makes sense.

yordis:
> I'd rather not have "Every Commanded Application, with default settings, is a time bomb."

@jvantuyl I would like to share some history about it based on HMB.

It was critical to have such "sharding" opportunities at multiple levels and understand how scalable the codebase could be as the application performance changes without having to rearchitect or change the system drastically. It was part of the decision-making and reasons to adopt an entirely new stack.

TL;DR about scaling the system over time: https://10consulting.com/2021/03/18/commanded-application-architecture/

You could disagree with it still, but everyone knew.

By design, it is good these level configurations exist.

---

Your concerns are valid, but it is a programmer problem rather than a technology problem. Commanded already allows you to do multi-app architecture styles.

jvantuyl:
I don't particularly disagree that users should learn their system and architect it accordingly.  And I'm a huge fan of systems that are flexible enough to handle multiple styles of multi-app architectures.  But...

Being "having sharp edges coming out of the box" isn't a programmer problem.  The [rule of least astonishment](https://code.hmbradley.io/core/backend/hmbradley/-/merge_requests/2179/commits) is pretty well established.  Defaults *can* surprise people.  That's entirely *possible*.  But it's still just poor ergonomics.

Consider the example aggregates in the docs: bank accounts, account balances.  Is there any real-world case where those should have infinite lifespans?  If the developer is thinking of this as a "database", is there any other database or cache that you can think of that doesn't automatically reclaim memory as it runs?

Don't just take it from me.  Take a look at blog posts.  One of the first ones you'll find is the excellent "Event Sourcing With Elixir" series by @saurdaukar.  Aggregate Lifespans show up in [the very last installment](https://blog.nootch.net/post/event-sourcing-with-elixir-part-7/) (part 7).  How are they introduced?  By an example of this exact behavior surprising the developer and them not noticing it until it was deployed and running for some time.  It's even in a section that's explicitly about the gotcha's of running Commanded in production.

[Here](https://medium.com/nexa-digital/my-journey-into-cqrs-and-event-sourcing-8f48a32b14ca) @rodrigobotti runs across it when adopting Commanded.  How does he introduce this?  "It might not be that apparent but our application has both memory and performance problems coming from our aggregates."  That's right.  It has a "memory problem" and "it might not be apparent".  This is not praise.

How about when @ChristianAlexander mentions it in [this blog post](https://christianalexander.com/2022/06/01/elixir-commanded-soft-delete/)?  He describes it as "That sounds like an OOM waiting to happen."

Or [here](https://q-the-hacker.com/post/how-we-reduced-memory-usage-in-an-application-built-with-elixir-and-commanded) where the entire blog post is about how this was killing an app in production?

Or search on Stack Overflow and you'll find that this has been [a problem from the beginning](https://stackoverflow.com/questions/53727392/avoiding-memory-leaks-with-commanded-in-an-aggregate-that-doesnt-produce-an-eve).  Note how the user describes it?  A "memory leak".  There's the clear connotation that this is so unexpected that it falls into a class of common bug.  And how did they find it?  Their app was running out of memory frequently.

In fact, if you Google for "elixir commanded lifespan", there isn't a single relevant result that doesn't cast this as a "gotcha" or a "problem".

If we're dead-set on keeping the "default behavior is to never free memory" thing, perhaps we can update the documentation to make it more obvious.  It's not even mentioned in the section on aggregates.  It's only briefly mentioned in at the very bottom of the section on Command Handlers.  Even then, it's pretty easy for a developer to miss that "By default an aggregate instance process will run indefinitely once started." means "Set this or you'll leak memory until you run out."

Or maybe we could alter `Commanded.Aggregates.DefaultLifespan` to emit a one-time warning the first time `after_command/1` is called?

Just something to make users think about this if they miss that part of the documentation?  Preferably before they have to explain to their boss why the tool they picked is crashing in production?

slashdotdash:
This PR doesn't address the legitimate concern raised by @jvantuyl that Commanded provides a default aggregate lifespan which keeps aggregate processes run indefinitely. I'm in agreement that it is OK for it be changed. We could be set to an appropriate timeout value, something like 24 hours?

Secondly, the default aggregate lifespan configuration should be associated with a Commanded Application, rather than being a global value associated with the `:commanded` app. Such global configuration is undesirable (e.g. difficult or impossible to test). We can support per Application configuration by using the `Commanded.Application.Config` module and with appropriate documentation added to the `Commanded.Application` module docs.

cdegroot:
If anyone is interested in this, then they should do a quick fix-up, otherwise I propose we close it for now (should be a topic we may want to revisit, though! It's useful to be able to set a default lifespan and fully agree with @slashdotdash' remarks)

cdegroot:
(I guess #548 supercedes this?

jwilger:
Closing per discussion.

PR #542, opened 2023-10-26
Use `Logger.warning` to fix deprecation warnings

janpieper:
When using commanded with Elixir 1.15, we currently see a lot deprecation warnings for `Logger.warn/1`:

```
warning: Logger.warn/1 is deprecated. Use Logger.warning/2 instead                                                                                                                                                   
  lib/commanded/middleware/consistency_guarantee.ex:47: Commanded.Middleware.ConsistencyGuarantee.after_dispatch/1
```

It should be fine to change `Logger.warn/1` to `Logger.warning/1` because the _new_ function already exists since Elixir 1.11.

FYI: `eventstore` already has already been updated to use `Logger.warning/1`: https://github.com/commanded/eventstore/pull/278

---

@slashdotdash What about this? :arrow_down_small: 

https://github.com/commanded/commanded/blob/e20fddd01f46a6668ce7360761c7136e5e3bc497/mix.exs#L10

and

https://github.com/commanded/commanded/blob/e20fddd01f46a6668ce7360761c7136e5e3bc497/README.md?plain=1#L22

Can we just increase these to at least 1.11? :thinking:

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ You can use Commanded with one of the following event stores for persistence:
 
 Please refer to the [CHANGELOG](CHANGELOG.md) for features, bug fixes, and any upgrade advice included for each release.
 
-Requires Erlang/OTP v21.0 and Elixir v1.9 or later.
+Requires Erlang/OTP v21.0 and Elixir v1.11 or later.
 
 ---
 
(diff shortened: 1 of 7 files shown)


slashdotdash:
@janpieper Let's bump the minimum Elixir version to v1.11

janpieper:
@slashdotdash Just pushed a new commit with increasing the minimum required elixir version to 1.11 :wink:

PR #547, opened 2024-01-10
Refactor router macros to limit compile-time dependencies

xpgdk:
Updated to apply cleanly

### DIFF
--- a/lib/commanded/commands/router.ex
+++ b/lib/commanded/commands/router.ex
@@ -293,16 +293,13 @@ defmodule Commanded.Commands.Router do
     aggregate_module =
       Macro.expand_literals(aggregate_module, %{__CALLER__ | function: {:dummy, 1}})
 
-    {opts, _bindings} = Code.eval_quoted(opts, [], __CALLER__)
-
-    Commanded.Commands.Router.register_identity(
-      __CALLER__,
-      __CALLER__.module,
-      aggregate_module,
-      opts
-    )
-
-    []
+    quote bind_quoted: [opts: opts, aggregate_module: aggregate_module] do
+      Commanded.Commands.Router.register_identity(
+        __MODULE__,
+        aggregate_module,
+        opts
+      )
+    end
   end
 
   @doc """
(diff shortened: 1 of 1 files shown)


lib/commanded/commands/router.ex
```diff
@@ -288,48 +290,19 @@ defmodule Commanded.Commands.Router do
...
-
-              invalid ->
-                raise "#{inspect(aggregate_module)} aggregate identity has an invalid `by` option: #{inspect(invalid)}"
-            end
-
-          prefix =
-            case Keyword.get(opts, :prefix) do
-              nil ->
-                nil
+    {opts, _bindings} = Code.eval_quoted(opts, [], __CALLER__)
```
slashdotdash, line 296: Hex docs for the [`Code.eval_quoted/3` function](https://hexdocs.pm/elixir/Code.html#eval_quoted/3) has the following warning which seems to be applicable here as the usage is within a macro.

> **Warning**: Calling this function inside a macro is considered bad practice as it will attempt to evaluate runtime values at compile time. Macro arguments are typically transformed by unquoting them into the returned quoted expressions (instead of evaluated).

I don't know what this code is supposed to be doing or why it is needed. The macros in this module are already difficult enough to understand without this added complexity. 

Can you help me (and other contributors to this code) to understand what is going on.

xpgdk: Well spotted. I think I managed to rewrite the macro to avoid using `eval_quoted`.
It was needed to properly handle the case where a function-reference is passed to the `by` option.

cdegroot (commented):
I have some trouble figuring out what compile-time dependencies are being limited and how. The changes are, well, I dunnow - the biggest seems to be refactoring the case statement but I'm not sure it'll make the code clearer; run-time, these couple of in-line case statements aren't gonna make a big dent on anything? @xpgdk care to elaborate a bit about the idea behind this PR?

lib/commanded/commands/router.ex
```diff
@@ -288,47 +290,15 @@ defmodule Commanded.Commands.Router do
...
-          @registered_identities Map.put(@registered_identities, aggregate_module,
-                                   by: by,
-                                   prefix: prefix
-                                 )
-
-        config ->
-          raise "#{inspect(aggregate_module)} aggregate has already been identified by: `#{inspect(Keyword.get(config, :by))}`"
-      end
+    aggregate_module =
+      Macro.expand_literals(aggregate_module, %{__CALLER__ | function: {:dummy, 1}})
```
yordis, line 294: What is `:dummy` function?

jwilger:
Closing due to age and lack of clarity around what this is improving. Feel free to reopen if you get back to this.

PR #550, opened 2024-01-23
Make before_reset/0 an explicit callback function

DilaksunB:


cdegroot (approved):
Change works for me. I like more explicitness in interfaces, etc.

yordis (approved):

drteeth (approved):

Issue #538, opened 2023-08-22
Process Manager state serialization breaks when using a custom TypeProvider with the JsonSerializer


jvantuyl:
## Summary

Process Managers don't serialize their state like anything else.  This breaks some assumptions about how the serializer and type providers interact.  It happens to work with the default config, but breaks with almost any custom `TypeProvider` implementation.

## Description

For aggregates, the `source_type` for serializing snapshots is taken from the `TypeProvider` [here](https://github.com/commanded/commanded/blob/v1.4.2/lib/commanded/snapshotting.ex#L49).

For events, the `source_type` for serializing snapshots is taken from the `TypeProvider` [here](https://github.com/commanded/commanded/blob/v1.4.2/lib/commanded/event/mapper.ex#L57).

This is all well and good.  However, for process managers the `source_type` for serializing snapshots is hardcoded to use the module name [here](https://github.com/commanded/commanded/blob/v1.4.2/lib/commanded/process_managers/process_manager_instance.ex#L538).

That becomes problematic when you're using a custom `TypeProvider` together with the default `JsonSerializer` implementation.  That's because it tries to look up the type from the `TypeProvider` [here](https://github.com/commanded/commanded/blob/v1.4.2/lib/commanded/serialization/json_serializer.ex#L26).

So long as you don't use the JSON serializer or don't use a custom `TypeProvider`, that works fine.  When you use both, you get some really exciting errors in serialization.

## Recommended Fix

Use the `TypeProvider` to resolve the type for serializing process managers.

I don't think this will break any existing installations as they're either not encountering it, currently broken anyways, or have worked around it in a way that won't break.

## Our Current Issue

We're using `OnePiece.Commanded.TypeProvider` (see [here](https://hexdocs.pm/one_piece_commanded/OnePiece.Commanded.TypeProvider.html)) to combine type providers from all of our domains.  It lets us explicitly specify the names used to reference types.  We use the `prefix` feature as well, which automatically adds a prefix to the names for all of the registered types in a given individual domain.

This is nice because it decouples the database type names from actual module names and, consequently, how our code is structured.  Since events are forever, we consider this a pretty big deal.  There's no cleaning up the naming later without some database surgery that we'd prefer not to do.

This bug forces us to not register a "decoupled" name and to not use the prefix feature for any process manager.  Using prefixes also changes the name even if we use the module name, so we end up having them all collecting in our top-level type provider.

[...]

PR #558, opened 2024-06-19
Use TypeProvider for process managers snapshot serialization

vheathen:
Addresses #538

### DIFF
--- a/lib/commanded/process_managers/process_manager_instance.ex
+++ b/lib/commanded/process_managers/process_manager_instance.ex
@@ -6,7 +6,7 @@ defmodule Commanded.ProcessManagers.ProcessManagerInstance do
   require Logger
 
   alias Commanded.{Application, EventStore, Telemetry}
-  alias Commanded.EventStore.{RecordedEvent, SnapshotData}
+  alias Commanded.EventStore.{RecordedEvent, SnapshotData, TypeProvider}
   alias Commanded.ProcessManagers.{FailureContext, ProcessRouter}
 
   defmodule State do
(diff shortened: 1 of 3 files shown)


drteeth:
I think this change makes sense when you look at it on it's own. While it seems like a no brainer for the greenfield case, if it is accepted, what will it do when people upgrade?

It seems like the for the default case, `Commanded.Serialization.ModuleNameTypeProvider` will not be affected, but for anyone using a custom TypeProvider, will it not break their ProcessManager state?

Is there something we could do to alert users of custom type providers that we're about to break the contract (for the better)?

drteeth (changes requested):
We need to consider how to roll this change out

vheathen:
@drteeth sorry for the long reaction.

I agree that we need to communicate about this. My initial thought was to mention it in the changelog, but I'm not sure it will be enough. 

Of course, it is possible to introduce a separate configuration parameter to enable `TypeProvider` for process managers and keep the default behaviour intact, but it looks a bit heavy.

drteeth:
While trying to think through a valid migration path, I think I've come to agree with @jvantuyl in that the only people this would affect are people that would have worked around it. I think we should just go ahead and make the change.

drteeth:
The only other thing I'd like to see is a test that fails with the existing behaviour and then is resolved by making the change you are recommending. The idea being that we don't break it again in the future.

vheathen:
The only reason I didn't include tests is that I didn't find any tests for the actual `TypeProvider` usage, only the unit tests for the `ModuleNameTypeProvider`.

But of course, it's always better to have tests. Please check the included one.

drteeth (approved):

cdegroot (approved):

PR #561, opened 2024-07-02
Fix Error Handling Example

TylerPachal:
The Error Handling example was slightly incorrect.  A `FailureContext` struct needs to be returned, but the example was only returning the `context` map from inside the `FailureContext`.

To test this was working I setup a very small application.

<details>
<summary>New documentation setup here</summary>

```elixir
defmodule Playground do
  defmodule Application do
    use Commanded.Application,
      otp_app: :playground,
      event_store: [adapter: Commanded.EventStore.Adapters.InMemory]
  end

  defmodule Handler do
    use Commanded.Event.Handler, application: Playground.Application, name: "Handler"
    alias Commanded.Event.FailureContext
    require Logger

    def handle(_event, _metadata), do: {:error, :failed}

    # --- START OF DOCUMENTATION CODE (this was simply copied into the docs) ---
    def error({:error, :failed}, event, %FailureContext{} = failure_context) do
      failure_context = record_failure(failure_context)

      case Map.get(failure_context.context, :failures) do
        too_many when too_many >= 3 ->
          # skip bad event after third failure
          Logger.warning("Skipping bad event, too many failures: " <> inspect(event))

          :skip

        _ ->
          # retry event, failure count is included in context map
          {:retry, failure_context}
      end
    end

    defp record_failure(failure_context) do
      context = Map.update(failure_context.context, :failures, 1, fn failures -> failures + 1 end)
      %FailureContext{failure_context | context: context}
    end
    # --- END OF DOCUMENTATION CODE ---

  end

  def test do
    {:ok, _} = Playground.Application.start_link()
    {:ok, pid} = Playground.Handler.start_link()

    events = [%Commanded.EventStore.RecordedEvent{}]
    send(pid, {:events, events})
  end
end
```

```
$ iex -S mix run

# Observe the "retrying failed event" messages when running the test:
iex>  Playground.test()
```

image

</details>

### DIFF
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -475,10 +475,10 @@ defmodule Commanded.Event.Handler do
           {:error, :failed}
         end
 
-        def error({:error, :failed}, %AnEvent{} = event, %FailureContext{context: context}) do
-          context = record_failure(context)
+        def error({:error, :failed}, %AnEvent{} = event, %FailureContext{} = failure_context) do
+          failure_context = record_failure(failure_context)
 
-          case Map.get(context, :failures) do
+          case Map.get(failure_context.context, :failures) do
             too_many when too_many >= 3 ->
               # skip bad event after third failure
               Logger.warning("Skipping bad event, too many failures: " <> inspect(event))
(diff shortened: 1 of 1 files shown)


lib/commanded/event/handler.ex
```diff
@@ -487,12 +487,13 @@ defmodule Commanded.Event.Handler do
...
-              {:retry, context}
+              {:retry, failure_context}
           end
         end
 
-        defp record_failure(context) do
-          Map.update(context, :failures, 1, fn failures -> failures + 1 end)
+        defp record_failure(failure_context) do
+          context = Map.update(failure_context.context, :failures, 1, fn failures -> failures + 1 end)
+          %FailureContext{failure_context | context: context}
```
TylerPachal, line 496: Annotation: Wrapping the `context` in the `FailureContext` is the main fix.  The other changes were just to make it more clear you need to return the `FailureContext`.

TylerPachal:
Closing this because the code can actually handle a plain map or a `%FailureContext{}`.  So this code change was not incorrect, but not necessary.

Issue #461, opened 2021-11-09
Snapshot Event Handler State


marpo60:
### Context

Event Handler State is provided in the init/1 callback and can be updated in each handle/2
https://hexdocs.pm/commanded/1.3.0/Commanded.Event.Handler.html#module-event-handler-state

This is pretty handy, but as this moment the state is ephemeral, which means that on server restart, there is a need to reconstruct the state.

It would be awesome if the EventHandler state could be persisted in the same fashion as the Process Manager state.

https://github.com/commanded/commanded/blob/1c8af07e4fdc415446b80482bf92f4f9c9873269/lib/commanded/process_managers/process_manager_instance.ex#L82-L84

https://github.com/commanded/commanded/blob/1c8af07e4fdc415446b80482bf92f4f9c9873269/lib/commanded/process_managers/process_manager_instance.ex#L243

If there is value in doing this I can start implementing this.

slashdotdash:
Yes, I agree this would be a useful opt-in feature for event handlers. A pull request would be very welcome, thanks!

PR #567, opened 2024-09-11
feat: snapshot event handler state

emisilvacab:
Snapshot the Event Handler state when `persistence: :permanent`

Configurable with the `persistence` option (`:ephemeral` by default)

Related issue #461 

Credits to @lucazulian 🙌

### DIFF
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -335,8 +335,8 @@ defmodule Commanded.Event.Handler do
   alias Commanded.Event.Handler
   alias Commanded.Event.Upcast
   alias Commanded.EventStore
-  alias Commanded.EventStore.SnapshotData
   alias Commanded.EventStore.RecordedEvent
+  alias Commanded.EventStore.SnapshotData
   alias Commanded.EventStore.Subscription
   alias Commanded.Subscriptions
   alias Commanded.Telemetry
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -166,8 +166,8 @@ defmodule Commanded.Event.Handler do
 
   An event handler can define and update state which is held in the `GenServer`
   process memory. It is passed to the `handle/2` function as part of the
-  metadata using the `:state` key. The state is transient and will be lost
-  whenever the process restarts.
+  metadata using the `:state` key. By default the state is transient and will be lost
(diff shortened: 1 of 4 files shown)


lib/commanded/event/handler.ex
```diff
@@ -1087,6 +1108,60 @@ defmodule Commanded.Event.Handler do
...
+        }
+
+      {:error, :snapshot_not_found} ->
+        state
+    end
+  end
+
+  defp read_state(state), do: state
+
+  defp persist_state(%Handler{persistence: :permanent} = state) do
```
cdegroot, line 1137: Snapshotting on every event smells like something that will severely limit the usability of this feature; there's a reason that Aggregate snapshotting is only between checkpoints. 

Not going to make a breaking point, if this is a useful feature as it is, that's fine, but I wonder how this thing would work when only snapshotting every X events or Y minutes...

cdegroot (approved):

yordis:
Would it be better to move this complexity to a Read Model? As any other projection?

lib/commanded/event/handler.ex
drteeth: I’m torn here. It seems like a useful feature but I’m concerned that we’re using the same mechanism that process managers do and there are well known issues with serialization there.

I also wonder (like Yordis) if you couldn’t just manually do this yourself in an event handler.

I don’t want to accept this PR at this point in time. At least not without getting to the bottom of process manager state issues, and that won’t be any time soon.

It is a good idea though @emisilvacab. Thank you for putting the work in.

jwilger: I'm in agreement with @cdegroot and @drteeth. A potentially better approach that already works would be to use an ecto projection (or otherwise create some handler-specific storage of state) so that IO within event handling is made explicit.

jwilger (changes requested):
@emisilvacab It's looking like we may not accept this particular change, but I want to make sure you have a chance to respond before we close this out. We'll close this PR on 10/25 if there is no further discussion.

lib/commanded/event/handler.ex
yordis: Unless there is a **strong** technical limitation here, I personally believe we should guide people to create read-models, including a read-model that may feel like a snapshot, make it explicit to the processing of the event

PR #570, opened 2024-09-28
chore: remove asdf file

yordis:
Upgrading the latest version `asdf` (devtool only), so we can take advantage of latest tooling

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ erl_crash.dump
 /priv/plts/*.plt.hash
 /bench/graphs
 /bench/snapshots
+.tool-versions
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,0 @@
-elixir 1.15.7-otp-25
-erlang 25.3.2.8


cdegroot:
I think it's useful to develop on the oldest supported Elixir and Erlang versions, but I also think that there's no formal support policy?

yordis:
> I think it's useful to develop on the oldest supported Elixir and Erlang versions

Based on my experience maintaining other major organizations, I prefer to keep the CI current to see the problems as soon as possible or take advantage of it and follow https://hexdocs.pm/elixir/compatibility-and-deprecations.html to keep the matrix alive.

I only add things to the matrix when needed. Still, I fundamentally understand that I prefer to put enough trust in Elixir and OTP itself;

drteeth:
Seems like most hex packages don't include .tool-versions leaving devs to take their own path here. I don't have a big preference either way.

It does seem like we could flesh out the test matrix though:
https://github.com/commanded/commanded/blob/master/.github/workflows/test.yml#L13

yordis:
Based on my experience, `.tool-versions` is kept up to date and used in for formatting, listing, and things like that, using the `version-file` in `setup-beam`;

While some use matrixes for tests.

drteeth:
This is not an academic point btw:
https://github.com/commanded/commanded/actions/runs/11319982776

yordis:
> This is not an academic point btw:

I do not understand 😅. I am not making an argument for removing the matrix.

drteeth:
Sorry, I just mean that we have actual problems introduced by newer versions of Elixir and OTP, this is to your point.

yordis:
Should I close this?

drteeth:
I can't decide what to do here. I'm tempted to remove it altogether or update it now that we have the matrix updated to cover the supported versions.

@cdegroot @jwilger ?

yordis:
removed! this only affects people local development that are using `asdf`

drteeth:
Add it to the git ignore file too?

yordis:
done

yordis:
ping

PR #594, opened 2024-10-07
registration: workaround compiler peculiarity

predrag-rakic:
This is workaround for the issue described in https://github.com/commanded/commanded-swarm-registry/issues/2#issuecomment-1121685703

### DIFF
--- a/lib/commanded/registration.ex
+++ b/lib/commanded/registration.ex
@@ -83,6 +83,22 @@ defmodule Commanded.Registration do
       @before_compile unquote(__MODULE__)
 
       alias unquote(__MODULE__)
+
+      # This is workaround for the issue described in
+      # https://github.com/commanded/commanded-swarm-registry/issues/2#issuecomment-1121685703
+      # In short, "handle_xxx" clauses defined in `__before_compile__/1` are not available in runtime
+      # unless these dummy clauses are present in `__using__/1`
+      @impl true
+      def handle_call(:will_not_be_called, _, _) do
+      end
+
+      @impl true
+      def handle_cast(:will_not_be_called, _) do
+      end
+
+      @impl true
+      def handle_info(:will_not_be_called, _) do
+      end
     end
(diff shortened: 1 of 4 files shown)


predrag-rakic:
One of the reasons for `Commanded.Registration` module existence is to enable Commanded to delegate functionality to another package. But because of strange Elixir behavior (described in [swarm-registry issue](https://github.com/commanded/commanded-swarm-registry/issues/2#issuecomment-1121685703)), delegation mechanism does not work as implementers of the module would expect it to work (in fact it does not work at all). 

To observe the issue execute [test provided in this PR](https://github.com/commanded/commanded/pull/594/files#diff-782f3cb4beb767baa9cb33af56e968978f2df0969cf7956bbe621dd13303a49d) without the change in [`Commanded.Registration`](https://github.com/commanded/commanded/pull/594/files#diff-3ed9ef60bf4d4e782e7cbc91445ac3e17b7ca83ac231b951be2c11dccbfa203eR86-R101).  You'll see that `handle_xxx` functions defined in [`HandleFunctionDelegationRegistry`](https://github.com/commanded/commanded/pull/594/files#diff-0b4f9f35d984dd201b89225ab3288e238d14b677876182914a5299e6de2d6a2eR37-R50) are not called because delegation mechanism  does not work.

In short, `Commanded.Registration` needs to be fixed and that cannot be done outside Commanded.

drteeth:
Ouch this is a gnarly one. Thanks for taking the time to dig in and try to find a solution.

I'm concerned about the comment in the swarm issue:
> @slashdotdash Any suggestions on to solve this correctly? Just adding the functions to the __using__ seem to break other parts of the application where the use Commanded.Registration is used.

Isn't that what you've done here? Maybe I'm misreading.

drteeth:
Thanks for the work on this one, I'm closing it in favour of #607 which is largely the same idea.

PR #595, opened 2024-10-14
Support OTP 26 and Elixir 1.17

drteeth:
* Adds Elixir 1.13, 1.14, 1.16, 1.17, and OTP 26, OTP 27 to the test matrix
* Fix issues found within.
* Fixes test breakage due to OTP 26's map key ordering change
* Fixes test breakage due to OTP 26's change in Gen* synchronous init failures
* Addresses unhandled `:unexpected_ack` error in `InMemory` adapter 

See https://www.erlang.org/news/164 for details on OTP 26 changes


drteeth:
@slashdotdash if you have time I would love your opinion on what I've done here https://github.com/commanded/commanded/pull/595/commits/0488f8805808be74d0d522ac1db5b8819e970d26

There were many errors logged in the test run around the InMemory adapter not handling `publish_events(state, subscription)` where `{:error, :unexpected_ack}` is passed as the 2nd argument when no in_flight events are found on a persistent subscription.

I think the options are to swallow the error and return the subscription unchanged as I have done, or to bubble the error all the way up to the caller.


What I have done appears to work, but I'm not super confident it is the right thing to do.

The breakage occurs on Elixir 1.17/OTP 27 and presents in `test/event/reset_event_handler_test.exs:41`. It is `BankAccountsHandler`'s subscription that fails to ack the event.

slashdotdash:
> @slashdotdash if you have time I would love your opinion on what I've done here [0488f88](https://github.com/commanded/commanded/commit/0488f8805808be74d0d522ac1db5b8819e970d26)

The in-memory event store adaptet is _only_ really supposed to be used for unit tests so what you have done looks OK to me.

cdegroot (approved):
Docs are explicit - the in-memory adapter is just there for testing. So whatever it takes :)

PR #48, opened 2017-03-26
Use Elixir 1.4 Registry for aggregates

patrickdet:
This addresses issue #38. The homegrown registry has been replaced by using the Elixir standard library Registry. Let me know if this looks alright.

PR #600, opened 2024-11-06
add missing timeout for aggregates (gen_server) callback

bougueil:
commanded 1.4.6
Linux 6.8.0-48-generic #48-Ubuntu SMP PREEMPT_DYNAMIC Fri Sep 27 14:04:52 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
erlang 27.0
elixir 1.17.1-otp-27

Some aggregates never finish despite a lifespan_timeout set to an integer.
 I chased all aggregates callbacks and complete the ones with a missing timeout.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-erlang 27.0
-elixir 1.17.1-otp-27
+elixir 1.15.7-otp-25
+erlang 25.3.2.8
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,2 +1,2 @@
-elixir 1.15.7-otp-25
-erlang 25.3.2.8
+erlang 27.0
+elixir 1.17.1-otp-27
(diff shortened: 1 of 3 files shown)


drteeth:
@bougueil is it possible to add a test case that this PR fixes?

bougueil:
Hi Benjamin,

short answer after digging my Logger traces.

What happen to my aggregate ?

The aggregate has a lifespan_timeout of 5mn
it receives in our app in this order :

   1. handle_call: *:execute_command* # sets correctly the timeout
2. handle_info a *:aggregate_state* msg -> destroys the timeout

Sorry it's not clear for me yet why the *:aggregate_state* is sent.

But nothing forbid to send *:aggregate_state* to an aggregate ?

Thanks for this great software,

Appreciate your help.

Renaud Mariana

drteeth:
Hmm that is a tricky one to test for sure. Thanks for looking into it.

I understand the problem as you've traced it, and it makes sense to me that we need to have all of the handle_call/cast/info callbacks keep the lifetime. I may see if I can write some tests for those.

What I don't understand yet is why the lifespan change is needed on the LocalRegistry. Can you help me understand that?

If it is needed by the LocalRegister, it is probably also needed by the global registry and any custom registries (Horde, Swarm, etc).

drteeth:
Closing in favour of #606

Issue #604, opened 2024-11-21
Implement exponential  backoff for retrying failed 


drteeth:
Many people crash their applications with a bug in an EventHandler that cascades up the supervision tree. The default behaviour is to return `{:stop, reason}` in the error callback.

For a poison pill event, or misconfigured handler, there is no way to make progress and you have to go out of your way to configure your app to avoid this.

@yordis blankets the application supervisor with:
```elixir
opts = [strategy: :one_for_one, name: Umbrella.Supervisor, max_restarts: 100, max_seconds: 5]
```
This results a ton of log-spam, but the application won't crash, and the developers have a reason to fix it quickly.

@cdegroot sets a "1 minute sleep" between restarts so as to avoid overwhelming the application, but still keeping the door open to transient errors being caught.

I wanted to default Event.Handlers to `restart: :temporary` to stop handlers at the first sign of trouble. 

@yordis rightly talked me out of that and proposed something like:
```elixir
defmodule MyCrashyHandler do
  use Commanded.Event.Handler,
    retry: :exponential,
    ...
```

This would give users an easy way to opt-in to a retry that will slowly backoff, giving o11y and APM tools plenty of opportunity to report the error and raise the alarm without overwhelming resources.

drteeth:
Addressed by https://github.com/commanded/commanded/pull/605

PR #605, opened 2024-11-21
Application-wide event handler error handling

drteeth:
Addresses #604 

## What

Introduce a mechanism for Commanded application level error handling strategies in general. Includes `:stop` and `:backoff` as supported strategies. Also includes a mechanism for custom error handling.

## Why
A classic stumbling block for newer Commanded user and people first going to production is that a single problem with an event, or handler, can result in their entire application crashing. 

We can't change the default behaviour without some nasty surprises, so this PR includes what is hopefully a reasonable strategy that can be opted into for all handlers in an application.


lib/commanded/event/handler.ex
```diff
@@ -626,12 +626,38 @@ defmodule Commanded.Event.Handler do
...
+    end
+  end
+
+  def handle_error(:stop, {:error, reason}, _failed_event, _failure_context) do
+    {:stop, reason}
+  end
+
+  def handle_error(:backoff, {:error, reason}, _event, %FailureContext{context: context}) do
+    context = Map.update(context, :failures, 1, fn failures -> failures + 1 end)
+    max_failures = Map.get(context, :max_failures, 2)
```
yordis, line 645: From the context, should we do `{:backoff, [max_failures: 2]` as well?

1. Check the Context
2. Check the compile-time value

lib/commanded/event/handler.ex
```diff
@@ -626,12 +626,38 @@ defmodule Commanded.Event.Handler do
...
+
+      defoverridable init: 1, after_start: 1, before_reset: 0, error: 3
+    end
+  end
+
+  def handle_error(:stop, {:error, reason}, _failed_event, _failure_context) do
+    {:stop, reason}
+  end
+
+  def handle_error(:backoff, {:error, reason}, _event, %FailureContext{context: context}) do
```
yordis, line 643:
```suggestion
  def handle_error(:random_backoff, {:error, reason}, _event, %FailureContext{context: context}) do
```
Should the name be a bit more precise?

lib/commanded/event/handler.ex
```diff
@@ -626,12 +626,38 @@ defmodule Commanded.Event.Handler do
...
+    context = Map.update(context, :failures, 1, fn failures -> failures + 1 end)
+    max_failures = Map.get(context, :max_failures, 2)
+
+    case Map.get(context, :failures) do
+      too_many when too_many >= max_failures ->
+        Logger.warning("Stopping #{__MODULE__} after too many failures.")
+        {:stop, reason}
+
+      failures ->
+        delay = trunc(failures ** 2) * 1000 + trunc(:rand.uniform() * 1000)
```
yordis, line 653:
```suggestion
        delay = trunc(failures ** 2) * 1000 + trunc(:rand.uniform() * 1000)
```

Should 1000 be configurable like `{:backoff, [find_better_name: 1000]`?

lib/commanded/event/handler.ex
drteeth: I've changed the implementation, but I think we'll put a few common ones in the box, and let you provide your own module, or override error/3 so you can do custom stuff. WDYT?

lib/commanded/event/error_handler.ex
```diff
@@ -0,0 +1,13 @@
...
+
+  @type failure_context :: map() | %FailureContext{context: map}
+  @type reason :: term()
+  @type delay :: non_neg_integer()
+
+  @callback handle_error({:error, reason}, struct(), %FailureContext{context: map}) ::
+              :skip
+              | {:retry, failure_context}
+              | {:retry, delay, failure_context}
+              | {:stop, reason}
```
yordis, line 12:
```suggestion
  @callback handle_error({:error, reason}, struct(), %FailureContext{context: map}, opts: keyword()) ::
              :skip
              | {:retry, failure_context}
              | {:retry, delay, failure_context}
              | {:stop, reason}
```

you will probably require some configuration for the strategy

lib/commanded/event/handler.ex
```diff
@@ -868,6 +872,35 @@ defmodule Commanded.Event.Handler do
...
+        Commanded.Event.ErrorHandler.Backoff
+
+      :grumpy_old_bastard ->
+        Commanded.Event.ErrorHandler.GrumpyOldBastard
+
+      module when is_atom(module) ->
+        # todo check this at macro time, why wait?
+        if function_exported?(module, :handle_error, 3) do
+          module
+        else
```
yordis, line 890: let it crash, HARD! I dont know, defensive programming drives me nuts

lib/commanded/event/handler.ex
```diff
@@ -868,6 +872,35 @@ defmodule Commanded.Event.Handler do
...
+  def extract_error_handler_module(opts) do
+    case Keyword.get(opts, :retry, :stop) do
+      :stop ->
+        Commanded.Event.ErrorHandler.Stop
+
+      :backoff ->
+        Commanded.Event.ErrorHandler.Backoff
+
+      :grumpy_old_bastard ->
+        Commanded.Event.ErrorHandler.GrumpyOldBastard
```
yordis, line 884: should it be just functions thou? instead of forcing to be a module

yordis (changes requested):

lib/commanded/event/handler.ex
drteeth: I don't think so, as the caller, I would just want to specify that I want it to backoff. I would either expect jitter to be taken care of for me, or I wouldn't even know what that is.

lib/commanded/event/handler.ex
drteeth: 1000 is just the conversion from seconds to ms here

lib/commanded/event/error_handler.ex
drteeth: I want to keep it simple, if you have advanced needs: specify a module or override on_error/3

lib/commanded/event/handler.ex
drteeth: I agree, this will crash at compile time

lib/commanded/event/handler.ex
drteeth: I like that, I'll try it.

lib/commanded/event/handler.ex
drteeth: Renamed `on_error:`

lib/commanded/event/handler.ex
drteeth: Done

lib/commanded/event/handler.ex
```diff
@@ -533,6 +534,20 @@ defmodule Commanded.Event.Handler do
...
+
+                   :backoff ->
+                     &ErrorHandler.backoff/3
+
+                   :grumpy_old_bastard ->
+                     &ErrorHandler.grump/3
+
+                   error_handler when is_function(error_handler, 3) ->
+                     error_handler
+                 end)
```
yordis, line 549: I rather have `retry_policy` + `error` callback that this to be honest, otherwise, `error` callback sounds good enough

I am not sure, I don't like the API somewhat like this

drteeth: I'm not super sure what you are saying here, but I agree that something is confusing here. Having an `error/3` callback... but also an `on_error` mechanism is redundant.

We need to keep the `error/3` callback...

yordis: We need to decide if the retry mechanism will trigger before it reaches the error callback.

Can we do that first?

drteeth: `error/3` is meant to be the place where you customize how you deal with errors in an event handler. That is an excellent option for customizing the behaviour of 1 event handler.

If we had no restrictions, I think making exponential backoff the default would be the best option. We can't do that because that would be a surprising and confusing breaking change.

I don't think we can separate retrying from the error callback, they are linked and I think that is appropriate. It would be nice if you could specify either the configuration option OR a custom callback, and get an error if you try to do both. I haven't found a good way of doing that yet.

We could walk back a step and only provide implementations functions for backoff and other strategies that you could call from `error/3`:

```elixir
@impl true
def error(_error, _event, _context) do
  backoff(context, max_retries: 42)
end
```
It's not as nice as the configuration you suggested, but much less tangled.

Finally, I wanted to make this a change to the default so that all of the handlers in a project would have this behaviour without having to remember to opt-in to it. What if we made a global or Commanded application level configuration that would apply to all of the handlers which would apply unless you implement `error/3`. This would signal your desire to override the configured behaviour.

Something like:
```elixir
# config/config.exs

config :commanded,
  default_error_handler: :backoff

# or

config MyApp.CommandedApp,
  default_error_handler: :backoff 

```

jwilger: > We can't do that because that would be a surprising and confusing breaking change.

Could we do it in a major version bump, though? I agree that would be a better out-of-the-box experience because "event handler error takes down entire app" is one of the biggest complaints I've encountered from folks who are new at working with OTP and don't yet understand how to deal with it on a supervision tree level.

jwilger: N.B. I wouldn't hold up *this* PR for that major version change discussion; that was more of an aside.

drteeth: I think if we do it this way, we can swap the default easily in a 2.0.

PR #607, opened 2024-11-30
Allow registration handle_call/cast callbacks to be called

drteeth:
Commanded modules ProcessRouter, Aggregate, and Event.Handler all `use` both `GenServer` and `Commanded.Registration` macros.

While the code as written with handle_call/cast/info callbacks allowed for Registration.handle_info/2 to be called by sending the process a message, the other two handlers were not reachable.

By moving those two into the `__using__` block, we get what we're after.

Should resolve https://github.com/commanded/commanded-swarm-registry/issues/2

Big thanks to @predrag-rakic for finding, reporting and proposing the fix.

### DIFF
--- a/lib/commanded/registration.ex
+++ b/lib/commanded/registration.ex
@@ -83,15 +83,7 @@ defmodule Commanded.Registration do
       @before_compile unquote(__MODULE__)
 
       alias unquote(__MODULE__)
-    end
-  end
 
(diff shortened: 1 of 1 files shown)


yordis (approved):
interesting

predrag-rakic:
This is clean solution. Thank you.

PR #616, opened 2025-02-08
feat: allow to set the event id

yordis:
## Context

I am working on a domain that must deal with Package Allocation, very similar to an Account Ledger, where I need to `Release` or `Commit` allocations.

They are workflows that may need to be retried due to a dual-write situation between appending to the EventStore and acknowledging the events (a Transfer saga).

## Research Done

I tried to rely on the Aggregate state to guarantee idempotency by saving an `idempotency_key` into the `MapSet` when a given Release or Allocation happens. Still, we ran into the issues of deciding how many IDs we should keep in memory or for how long we would keep it in memory, snapshotting the state, etc.

We could make assumptions and trade-offs, but it felt odd.

Therefore, I asked **Greg Youn** and **Oskar Dudycz** for some advice, and both of them suggested generating a deterministic Event ID that my Event Store would guarantee idempotency with.


yordis:
It seems that there is a non-deterministic test failing ... I can try to follow up on that in different PR

.dialyzer_ignore.exs
```diff
@@ -47,5 +47,27 @@
   {":0:unknown_function Function Commanded.Serialization.JsonDecoder.Reference.__impl__/1 does not exist.",
    :unknown_function},
   {":0:unknown_function Function Commanded.Serialization.JsonDecoder.Tuple.__impl__/1 does not exist.",
+   :unknown_function},
```
yordis, line 50: This is required by `Erlang v23.x` 😭

yordis:
Screenshot 2025-02-12 at 3 47 31 PM

yordis:
@drteeth @jwilger @cdegroot any updates over here? Is there anything I could do at the moment?

PR #636, opened 2025-07-17
[KA-3069] Make Aggregate.take_snapshot/3 a blocking call

satom99:


yordis:
What is the motivation to make the function a blocking call?

yordis:
I just realized that I may made the wrong assumptions. Just to confirm, is this function part of the execution command code path?

satom99:
I have changed the `handle_cast/2` clause into a `handle_info/2` and changed the `GenServer.cast` call into a call to `Kernel.send/2`. I believe it is a lot clearer now that that code path is intended for internal usage within the aggregate GenServer itself 🙂

lib/commanded/aggregates/aggregate.ex
```diff
@@ -396,6 +399,24 @@ defmodule Commanded.Aggregates.Aggregate do
     end
   end
 
+  @doc false
+  @impl GenServer
+  def handle_info({:take_snapshot, lifespan_timeout}, %Aggregate{} = state) do
+    state = %Aggregate{state | lifespan_timeout: lifespan_timeout}
+
+    Logger.debug(describe(state) <> " recording snapshot")
```
drteeth, line 407: I think it makes sense to leave the logger statement in `do_take_snapshot/1`

lib/commanded/aggregates/aggregate.ex
```diff
@@ -604,20 +625,10 @@ defmodule Commanded.Aggregates.Aggregate do
...
-          %Aggregate{state | snapshotting: snapshotting}
-
-        {:error, error} ->
-          Logger.warning(describe(state) <> " snapshot failed due to: " <> inspect(error))
-
-          state
-      end
-
-    noreply_with_lifespan(state)
+    with {:ok, snapshotting} <-
```
drteeth, line 628: I like that you left `noreply_with_lifespan/1` out of here.

I think the single-case `with` should go back to being a `case` here.

drteeth (commented):
I like this, thanks for submitting. Also thanks to @yordis for helping make sense of it.

lib/commanded/aggregates/aggregate.ex
satom99: I added a `case` back in there together with the logs in cfef270fb330ab3ba452d6a4d710aa88f4a46af7 🙇

lib/commanded/aggregates/aggregate.ex
satom99: Agreed, changed in cfef270fb330ab3ba452d6a4d710aa88f4a46af7!

drteeth (approved):

satom99:
@drteeth done!

cdegroot (approved):

PR #637, opened 2025-07-17
[KA-3087] Add :dispatch_opts to ProcessManager

satom99:
This would allow specifying a custom dispatch timeout for each ProcessManager rather than having to fiddle with the global timeout.

### DIFF
--- a/lib/commanded/process_managers/process_manager.ex
+++ b/lib/commanded/process_managers/process_manager.ex
@@ -631,7 +631,8 @@ defmodule Commanded.ProcessManagers.ProcessManager do
     :subscribe_to,
     :subscription_opts,
     :event_timeout,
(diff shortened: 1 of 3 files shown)


drteeth (commented):
I'd like to know about the use case you are trying to support here. I get that you'd like to configure a dispatch timeout per progress manager, but I'm wondering why the solution is to change the timeout. Is this because you are using strong consistency?

If that does make sense, I'm not sure the approach of exposing all dispatch options to ProcessManagers makes sense. I think following the example of passing specific options in here lets us consider each one as it relates to ProcessManagers.

yordis:
Another section for my blog post to make sure I do not forget to mention "I'm not sure the approach of exposing all dispatch options to ProcessManagers makes sense"

PR #641, opened 2025-11-15
✨ add timeout option to batched event handler

rparcus:
Adds a `batch_timeout` option to event handlers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## vNext (Unreleased)
+
+### Enhancements
+
+* Add `batch_timeout` option for time-based batch flushing
+  - Batches now flush on size OR time, whichever comes first
+  - Solves unbounded latency problem for low-volume event streams
+  - Defaults to `:infinity` for full backwards compatibility
+  - Follows Broadway/GenStage patterns for familiar API
+  - See event handler guide for usage examples and timeout recommendations
+
+### Bug fixes
+
+* Fix race condition in batch processing where timer could fire during batch handling
+  - Buffer and timer now cleared before processing to prevent stale state
+
 ## v1.4.9
 
 ### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,6 @@
   - Follows Broadway/GenStage patterns for familiar API
   - See event handler guide for usage examples and timeout recommendations
 
-### Bug fixes
-
-* Fix race condition in batch processing where timer could fire during batch handling
-  - Buffer and timer now cleared before processing to prevent stale state
-
 ## v1.4.9
 
 ### Enhancements
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -862,7 +862,13 @@ defmodule Commanded.Event.Handler do
     end
 
     {batch_size, config} = Keyword.pop(config, :batch_size)
-    {batch_timeout, config} = Keyword.pop(config, :batch_timeout, :infinity)
+    {batch_timeout_opt, config} = Keyword.pop(config, :batch_timeout, :__no_batch_timeout__)
+    batch_timeout =
+      case batch_timeout_opt do
+        :__no_batch_timeout__ -> :infinity
+        value -> value
+      end
+    batch_timeout_provided? = batch_timeout_opt != :__no_batch_timeout__
 
     # Validate batch_size
     unless is_nil(batch_size) or (is_integer(batch_size) and batch_size > 0) do
@@ -883,6 +889,12 @@ defmodule Commanded.Event.Handler do
     config =
       case batch_size do
         nil ->
+          if batch_timeout_provided? do
+            raise ArgumentError,
+                  inspect(module) <>
+                    " :batch_timeout requires :batch_size. Remove the timeout or configure batching."
+          end
+
           # Delegate to `handle_event/2` when `batch_size` is not specified
           config
           |> Keyword.put(:handler_callback, :event)
@@ -1208,8 +1220,9 @@ defmodule Commanded.Event.Handler do
 
         # Check if we should flush based on size
         if length(new_buffer) >= batch_size do
-          flush_batch_buffer(state, :size)
+          state
           |> cancel_batch_timer()
+          |> flush_batch_buffer(:size)
         else
           state
         end
@@ -1254,8 +1267,22 @@ defmodule Commanded.Event.Handler do
   defp cancel_batch_timer(%Handler{batch_timer_ref: nil} = state), do: state
 
   defp cancel_batch_timer(%Handler{batch_timer_ref: ref} = state) do
-    Process.cancel_timer(ref)
-    %Handler{state | batch_timer_ref: nil}
+    case Process.cancel_timer(ref) do
+      false ->
+        drain_flush_batch_timeout_message()
+        %Handler{state | batch_timer_ref: nil}
+
+      _remaining ->
+        %Handler{state | batch_timer_ref: nil}
+    end
+  end
+
+  defp drain_flush_batch_timeout_message do
+    receive do
+      :flush_batch_timeout -> :ok
+    after
+      0 -> :ok
+    end
   end
 
   defp handle_event(event, context \\ %{}, handler)
--- a/lib/commanded/event/handler.ex
+++ b/lib/commanded/event/handler.ex
@@ -849,6 +849,13 @@ defmodule Commanded.Event.Handler do
             "both `:concurrency` and `:batch_size` are specified, this is not yet supported. Please choose one or the other."
     end
 
+    # batch_timeout requires batch_size
+    if Keyword.has_key?(config, :batch_timeout) and not Keyword.has_key?(config, :batch_size) do
+      raise ArgumentError,
+            inspect(module) <>
+              " :batch_timeout requires :batch_size. Remove the timeout or configure batching."
+    end
+
     {application, config} = Keyword.pop(config, :application)
 
     unless application do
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
...
+  - Solves unbounded latency problem for low-volume event streams
+  - Defaults to `:infinity` for full backwards compatibility
+  - Follows Broadway/GenStage patterns for familiar API
+  - See event handler guide for usage examples and timeout recommendations
+
+### Bug fixes
+
+* Fix race condition in batch processing where timer could fire during batch handling
+  - Buffer and timer now cleared before processing to prevent stale state
+
```
rparcus:
```suggestion
```

lib/commanded/event/handler.ex
```diff
@@ -776,20 +851,54 @@ defmodule Commanded.Event.Handler do
...
+    end
 
     config =
       case batch_size do
         nil ->
+          if batch_timeout_provided? do
+            raise ArgumentError,
+                  inspect(module) <>
+                    " :batch_timeout requires :batch_size. Remove the timeout or configure batching."
+          end
```
TylerPachal: I think this check makes more sense with the checks up above like "both `:concurrency` and `:batch_size` are specified..." at the beginning of this function.  

A bonus of moving the checks is that I think you would no longer need the logic surrounding `__no_batch_timeout__`.

lib/commanded/event/handler.ex
```diff
@@ -1040,6 +1176,104 @@ defmodule Commanded.Event.Handler do
...
+    } = state
+
+    # Upcast events before buffering
+    events = Upcast.upcast_event_stream(events, additional_metadata: %{application: application})
+
+    # Determine processing strategy based on configuration
+    cond do
+      # No batch_size configured - process immediately
+      is_nil(batch_size) ->
+        handle_batch(events, %{flush_reason: :immediate}, state)
```
TylerPachal: Is this case actually possible? If we are in `:batch` callback mode doesn't there have to be a `batch_size`  configured upstream from here?  Enforced by this code: https://github.com/commanded/commanded/blob/master/lib/commanded/event/handler.ex#L782-L784

If this case is not possible, and there are actually only two possible cases here, I think this code would be a little easier to follow if you moved things up into the `handle_info` above and had all three code paths directly in the `case` statement:

```elixir
try do
  events = Upcast.upcast_event_stream(events, additional_metadata: %{application: application})

  state =
    case {callback, state.batch_timeout) do
      {:event, _} ->
        # Non-batched: process immediately
        Enum.reduce(events, state, &handle_event/2)

      # Batched with no timeout configured
      {:batch, nil} ->
        handle_batch(events, state)

      # Batched with timeout
      {:batch, batch_timeout} ->
        buffer_and_maybe_flush(events, state)
    end
  end

# ...

defp buffer_and_maybe_flush(events, %Handler{} = state) do
  %Handler{
    batch_buffer: buffer,
    batch_size: batch_size,
    batch_timeout: batch_timeout
  } = state

  current_buffer = buffer || []
  new_buffer = current_buffer ++ events
  state = %Handler{state | batch_buffer: new_buffer}

  # Start timer if this is first event in batch
  state = maybe_start_batch_timer(state)

  # Check if we should flush based on size
  if length(new_buffer) >= batch_size do
    state
    |> cancel_batch_timer()
    |> flush_batch_buffer(:size)
  else
    state
  end
end
```

I think this makes more sense because the `buffer_and_maybe_flush` function now is clearly only on the codepath with the timer.  It was a little confusing about why all `:batch` paths would go through the buffer + flush mechanism.

lib/commanded/event/handler.ex
```diff
@@ -1040,6 +1176,104 @@ defmodule Commanded.Event.Handler do
...
+  defp buffer_and_maybe_flush(events, %Handler{} = state) do
+    %Handler{
+      application: application,
+      batch_buffer: buffer,
+      batch_size: batch_size,
+      batch_timeout: batch_timeout
+    } = state
+
+    # Upcast events before buffering
+    events = Upcast.upcast_event_stream(events, additional_metadata: %{application: application})
```
TylerPachal: If this is being called here, and in the `:event` codepath of `handle_info({:events, events}, state)` [here](https://github.com/commanded/commanded/pull/641/changes#diff-64db9844d93a959af4268d5f685db40f98dfbb5781bfa60b41bd8755e8d5b984R1081-R1084) then I think its simpler for it to reside only in the `handle_info({:events, events}, state)` function (at the top of the `try` block).

lib/commanded/event/handler.ex
```diff
@@ -776,20 +869,42 @@ defmodule Commanded.Event.Handler do
     end
 
     {batch_size, config} = Keyword.pop(config, :batch_size)
+    {batch_timeout, config} = Keyword.pop(config, :batch_timeout, :infinity)
+
+    # Validate batch_size
+    unless is_nil(batch_size) or (is_integer(batch_size) and batch_size > 0) do
```
drteeth, line 863: `unless` was deprecated in Elixir 1.18, can you refactor this to use `if` please?

lib/commanded/event/handler.ex
```diff
@@ -776,20 +869,42 @@ defmodule Commanded.Event.Handler do
...
+      raise ArgumentError,
+            inspect(module) <>
+              " :batch_size must be nil or positive integer, got: " <>
+              inspect(batch_size)
+    end
+
+    # Validate batch_timeout
+    unless batch_timeout == :infinity or (is_integer(batch_timeout) and batch_timeout > 0) do
```
drteeth, line 871: `unless` was deprecated in Elixir 1.18, can you refactor this to use `if` please?

lib/commanded/event/handler.ex
```diff
@@ -1040,6 +1191,88 @@ defmodule Commanded.Event.Handler do
...
+  defp flush_batch_buffer(%Handler{batch_buffer: nil} = state, _flush_reason), do: state
+
+  defp flush_batch_buffer(%Handler{batch_buffer: buffer} = state, flush_reason) do
+    Logger.debug(
+      describe(state) <> " flushing batch of #{length(buffer)} event(s) due to #{flush_reason}"
+    )
+
+    # Clear buffer and timer BEFORE processing to prevent race condition
+    # If timer fires during batch processing, it will see empty buffer
+    state = %Handler{state | batch_buffer: [], batch_timer_ref: nil}
```
drteeth, line 1219: Setting `batch_timer_ref` to nil here can't prevent a race condition can it?. You are cancelling the timer, draining any flush messages that made it through, and setting the timer ref to nil in `cancel_batch_timer/1` before this call.

lib/commanded/event/handler.ex
```diff
@@ -1040,6 +1191,88 @@ defmodule Commanded.Event.Handler do
...
+
+  defp maybe_start_batch_timer(state), do: state
+
+  # Cancel batch timer if running
+  defp cancel_batch_timer(%Handler{batch_timer_ref: nil} = state), do: state
+
+  defp cancel_batch_timer(%Handler{batch_timer_ref: ref} = state) do
+    case Process.cancel_timer(ref) do
+      false ->
+        drain_flush_batch_timeout_message()
```
drteeth, line 1248: Can you talk a bit about what this is for?

lib/commanded/event/handler.ex
```diff
@@ -952,21 +1076,32 @@ defmodule Commanded.Event.Handler do
...
+            Enum.reduce(events, state, &handle_event/2)
+
+          {:batch, timeout} when timeout in [:infinity, nil] ->
+            # Batched without timeout: process immediately (EventStore handles batching)
+            handle_batch(events, %{flush_reason: :immediate}, state)
+
+          {:batch, _timeout} ->
+            # Batched with timeout: buffer events and flush on size or timeout
+            buffer_and_maybe_flush(events, state)
+        end
```
drteeth, line 1092: Love this.

drteeth (commented):
I really like the idea of this change. Thanks for putting the time in. There are a few things I'm unclear about still, but we'll get there.

lib/commanded/event/handler.ex
drteeth: NVM I see now

lib/commanded/event/handler.ex
```diff
@@ -1040,6 +1191,88 @@ defmodule Commanded.Event.Handler do
...
+        %Handler{state | batch_timer_ref: nil}
+
+      _remaining ->
+        %Handler{state | batch_timer_ref: nil}
+    end
+  end
+
+  defp drain_flush_batch_timeout_message do
+    receive do
+      :flush_batch_timeout -> :ok
```
drteeth, line 1258: What happens if the next message isn't a flush messge? I think we'll crash here no? And if you do manage to change the pattern match, then we've taken the message out of the mailbox and we have to process it it, whatever it is?

Am I wrong?

TylerPachal:
@rparcus Could you explain a little bit what behavior you are trying to add/change with this PR?

From what I can see, if I set `batch_size: 100` and dispatch a command, my event handler sees that event immediately (i.e. it does not wait for another 99 events to "fill" the batch).  

Are you seeing different behavior than that?

rparcus:
> @rparcus Could you explain a little bit what behavior you are trying to add/change with this PR?
> 
> From what I can see, if I set `batch_size: 100` and dispatch a command, my event handler sees that event immediately (i.e. it does not wait for another 99 events to "fill" the batch).
> 
> Are you seeing different behavior than that?

You're right, `batch_size` alone does not cause accumulation/holdup of events. I tested this again, against the real PostgreSQL EventStore: with `batch_size: 100`, dispatching a single command results in `handle_batch/1` being called immediately with that one event. `batch_size` sets the subscription's `buffer_size` (in-flight window) only, so real batches only form during catch-up replay or under back-pressure. Found [the old issue](https://github.com/commanded/commanded/issues/261) describing this actually, handle_batch in most cases is nothing more than a normal handler wrapped in a list as it "does its thing" only during catchup.

With this change, the `batch_timeout` option adds something that doesn't exist in commanded today: time-based buffering at the handler level. When configured, events are accumulated in the handler process and flushed when either `batch_size` events have collected OR `batch_timeout` milliseconds have elapsed. This enables use cases like batching projection writes during steady-state live operation, not just during catch-up. I have updated the module docs to make this distinction more evident.

lib/commanded/event/handler.ex
rparcus: The `after 0` clause below makes this a "non-blocking selective receive", meaning that it only takes `:flush_batch_timeout` from the mailbox if one is already there, otherwise returns immediately. 

It never blocks and never consumes other messages. This pattern is used for draining a specific message after cancelling a timer (`Process.cancel_timer/1` returns false when the timer already fired, meaning the message may be in the mailbox).

drteeth: Crazy. I had no idea you could do that, thanks for showing me.

drteeth (approved):

drteeth:
@rparcus sorry for the delay. If you still want this, resolve the conflicts and I'll merge it right away

drteeth:
Rebased and closed as https://github.com/commanded/commanded/pull/653

PR #644, opened 2026-02-04
Improve Router Registration Error

### LINKED ISSUE #422: Dispatch without :to option gives an unexpected error message
I was creating a test without a `:to` option, and it blew up in a non-obvious way.

```elixir
== Compilation error in file test/process_managers/process_manager_telemetry_test.exs ==
** (KeyError) key :aggregate not found in: [application: nil, consistency: :eventual, returning: false, timeout: 5000, lifespan: Commanded.Aggregates.DefaultLifespan, metadata: %{}, retry_attempts: 10]
    (elixir 1.10.4) lib/keyword.ex:399: Keyword.fetch!/2
    (stdlib 3.11.2) erl_eval.erl:680: :erl_eval.do_apply/6
    (stdlib 3.11.2) erl_eval.erl:888: :erl_eval.expr_list/6
    (stdlib 3.11.2) erl_eval.erl:411: :erl_eval.expr/5
    (stdlib 3.11.2) erl_eval.erl:126: :erl_eval.exprs/5
    (elixir 1.10.4) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
    (stdlib 3.11.2) erl_eval.erl:680: :erl_eval.do_apply/6
```

Error is thrown [here](https://github.com/commanded/commanded/blob/v1.2.0/lib/commanded/commands/router.ex#L497), but the macro could raise a helpful error message instead

TylerPachal:
If you forgot to include the required `:to` option in `dispatch(...)` during your router registration you would get an unhelpful error (as noted in #422).  This PR fixes that, so that a more helpful error is returned.

Additionally, while trying to replicate this I found another bug with the registration arguments: if the `:to` option was not the first option in the list, things would break.   I've addressed that bug here as well because it was in the same part of the code.

For this PR, the first commit was adding the new tests (which failed), and the second commit is the changes to get the tests passing.

Closes #422

### DIFF
--- a/lib/commanded/commands/router.ex
+++ b/lib/commanded/commands/router.ex
@@ -359,7 +359,7 @@ defmodule Commanded.Commands.Router do
 
   """
   defmacro dispatch(command_module_or_modules, opts) do
-    opts = parse_opts(opts, [])
+    opts = parse_opts(opts)
 
     for command_module <- List.wrap(command_module_or_modules) do
       quote do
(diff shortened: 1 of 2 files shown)


lib/commanded/commands/router.ex
```diff
@@ -666,28 +666,29 @@ defmodule Commanded.Commands.Router do
     :consistency
   ]
 
-  defp parse_opts([{:to, aggregate_or_handler} | opts], result) do
-    case Keyword.pop(opts, :aggregate) do
```
TylerPachal, line 670: Annotation: This was problematic.  This assume that the `:to` was always before the `:aggregate`, and would break when that was not the case.  There was also a funny interaction with the `:function` option because of how it was set to either `:execute` or `:handle` (so again, depending on order, you could lose a user-provide value for `:function`).

The fix is to pull out the options we need so they can be compared together, instead of looping over them sequentially.

lib/commanded/commands/router.ex
```diff
@@ -666,28 +666,29 @@ defmodule Commanded.Commands.Router do
...
+      """
     end
-  end
 
-  defp parse_opts([{param, value} | opts], result) when param in @register_params do
-    parse_opts(opts, [{param, value} | result])
-  end
+    case {to, aggregate} do
+      {nil, _aggregate} ->
+        raise "dispatch missing required parameter: :to"
```
TylerPachal, line 683: Annotation: The `raise` is the actual improvement to #644.  See the unit test for an example.

drteeth (approved):

PR #647, opened 2026-02-18
Support :selector option for InMemory adapter

TylerPachal:
The `:selector` option is supported by the "real" eventstore, but not by the InMemory eventstore. This is causing a lot of my unit tests to fail as event handlers are recieving event types that they do not receive in non-test environments.

I tested this in my own project, I can now swap the InMemory adapter for the Postgres implementation and all of my test will still pass 👍

### DIFF
--- a/lib/commanded/event_store/adapters/in_memory.ex
+++ b/lib/commanded/event_store/adapters/in_memory.ex
@@ -95,6 +95,7 @@ defmodule Commanded.EventStore.Adapters.InMemory do
       concurrency_limit: Keyword.get(opts, :concurrency_limit),
       name: subscription_name,
       partition_by: Keyword.get(opts, :partition_by),
+      selector: Keyword.get(opts, :selector),
       start_from: start_from,
       stream_uuid: stream_uuid
     }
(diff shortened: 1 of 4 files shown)


lib/commanded/event_store/adapters/in_memory.ex
```diff
@@ -483,16 +484,31 @@ defmodule Commanded.EventStore.Adapters.InMemory do
...
   end
 
+  defp selected?(event, %PersistentSubscription{selector: selector})
+       when is_function(selector, 1) do
+    selector.(event)
+  end
+
+  defp selected?(_event, %PersistentSubscription{}) do
+    true
+  end
```
TylerPachal, line 513: Annotation: I pretty much copied this code from the real EventStore.

test/event_store/support/subscription_test_case.ex
```diff
@@ -208,6 +208,30 @@ defmodule Commanded.EventStore.SubscriptionTestCase do
         refute_receive {:events, _received_events}
       end
 
+      test "should skip events that do not match the :selector", %{
```
TylerPachal, line 211: Annotation: I am not sure if this is the most appropriate place to put this test or not. This file is a giant macro which is meant to support different event stores. However, it is only actually used for the InMemory event store. The selector option may not be something supported by all event stores, so that's why I am not sure if i should place these tests here. On the other hand, this is where all of the subscription tests are.

test/event_store/support/subscription_test_case.ex
```diff
@@ -830,6 +881,24 @@ defmodule Commanded.EventStore.SubscriptionTestCase do
       end)
     end
 
+    defp receive_events(event_store, event_store_meta, subscription, count) do
```
TylerPachal, line 884: Annotation: There was already a assert_receive_events but for my tests I thought it would be safer to receive all of the events, then assert on them as a collection.

lib/commanded/event_store/adapters/in_memory.ex
```diff
@@ -478,21 +479,39 @@ defmodule Commanded.EventStore.Adapters.InMemory do
...
-        unseen_event =
-          deserialize(state, unseen_event) |> set_event_number_from_version(stream_uuid)
+    unseen_event =
+      with %RecordedEvent{} = event <- Enum.at(events, position) do
+        state
+        |> deserialize(event)
+        |> set_event_number_from_version(stream_uuid)
+      end
 
+    cond do
```
TylerPachal, line 489: Annotation: My original implementation triggered the `Credo.Check.Refactor.Nesting` check and failed because I had three levels of nesting (only two are permitted by default).  To satisfy credo I needed to move this code around a bit more than I had inteneded, and introduce the `cond` statement instead of nested conditional statements.

lib/commanded/event_store/adapters/in_memory.ex
```diff
@@ -478,21 +479,39 @@ defmodule Commanded.EventStore.Adapters.InMemory do
...
           {:ok, subscription} -> publish_events(state, subscription)
           {:error, :no_subscriber_available} -> subscription
         end
 
-      nil ->
-        subscription
+      unseen_event ->
+        %RecordedEvent{event_number: event_number} = unseen_event
+        subscription = PersistentSubscription.checkpoint(subscription, event_number)
+        publish_events(state, subscription)
```
TylerPachal, line 502: Annotation: If the event does not match the selector, do not publish it, but still record the checkpoint on the subscription.

drteeth (approved):

PR #648, opened 2026-02-23
Fix Warnings for Elixir 1.19

TylerPachal:
Mimics the changes in the PR for the EventStore: https://github.com/commanded/eventstore/pull/319

- Set the minimum version of Elixir to 1.15
- Add Elixir 1.19 to Github Actions test matrix and fix warnings for 1.19
- Fix formatting warning

### DIFF
--- a/lib/commanded/aggregates/aggregate.ex
+++ b/lib/commanded/aggregates/aggregate.ex
@@ -348,7 +348,7 @@ defmodule Commanded.Aggregates.Aggregate do
         # be taken. When it finishes, it will set the timeout.
         {:reply, formatted_reply, state}
       else
-        state = %Aggregate{state | lifespan_timeout: lifespan_timeout}
+        state = %{state | lifespan_timeout: lifespan_timeout}
 
         reply_with_lifespan(formatted_reply, state)
       end
--- a/mix.exs
+++ b/mix.exs
@@ -7,7 +7,7 @@ defmodule Commanded.Mixfile do
     [
       app: :commanded,
       version: @version,
-      elixir: "~> 1.12",
+      elixir: "~> 1.15",
       elixirc_paths: elixirc_paths(Mix.env()),
       deps: deps(),
       description: description(),
--- a/test/process_managers/support/error/error_aggregate.ex
+++ b/test/process_managers/support/error/error_aggregate.ex
@@ -98,7 +98,7 @@ defmodule Commanded.ProcessManagers.ErrorAggregate do
     do: {:error, :failed}
 
   def execute(%ErrorAggregate{}, %ContinueProcess{process_uuid: process_uuid, reply_to: reply_to}),
-    do: %ProcessContinued{process_uuid: process_uuid, reply_to: reply_to}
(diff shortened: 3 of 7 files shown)


lib/commanded/aggregates/aggregate.ex
```diff
@@ -348,7 +348,7 @@ defmodule Commanded.Aggregates.Aggregate do
         # be taken. When it finishes, it will set the timeout.
         {:reply, formatted_reply, state}
       else
-        state = %Aggregate{state | lifespan_timeout: lifespan_timeout}
```
TylerPachal, line 351: Annotation: In Elixir 1.19 this is a compiler warning now.  The struct type is already known from a pattern match up above.

test/process_managers/support/error/error_aggregate.ex
```diff
@@ -94,11 +94,13 @@ defmodule Commanded.ProcessManagers.ErrorAggregate do
     raise message
   end
 
-  def execute(%ErrorAggregate{}, %AttemptProcess{}),
-    do: {:error, :failed}
+  def execute(%ErrorAggregate{}, %AttemptProcess{}) do
+    {:error, :failed}
+  end
 
-  def execute(%ErrorAggregate{}, %ContinueProcess{process_uuid: process_uuid, reply_to: reply_to}),
```
TylerPachal, line 100: Annotation: The formatter for 1.19 was treating this different than the other versions, so it would be valid for 1.19 but not the others. I wrestled with it for a bit before deciding it was easier to change it to a regular `do` instead of the single-line `, do:`

TylerPachal:
@cdegroot @drteeth The Commanded repository does not have a `.tool-versions` file but the Eventstore repository does have one.  Do we want to include one here to help standardize things?   Are we all using `asdf`?

test/process_managers/support/error/error_aggregate.ex
drteeth: I prefer regular `do` personally

mix.exs
```diff
@@ -7,7 +7,7 @@ defmodule Commanded.Mixfile do
     [
       app: :commanded,
       version: @version,
-      elixir: "~> 1.12",
+      elixir: "~> 1.15",
       elixirc_paths: elixirc_paths(Mix.env()),
```
drteeth, line 11: Is this required? I know it matches the supported version matrix, but we're chopping off 3 versions of (deprecated) Elixir that someone may be stuck with.

TylerPachal: Oh, haha, I was doing this because the same change was being made for the eventstore: https://github.com/commanded/eventstore/pull/319/changes#r2822309888

Keeping the older version is fine with me, there are no changes in this PR which require 1.15.  I can also undo the changes for the test matrix.

drteeth: Maybe we should just go ahead with 1.15. We can go backwards if anyone really needs, or they can in a fork.

drteeth (approved):

mix.exs
TylerPachal: 1.15 was released in June of 2023.  So I think this is reasonable.  And yes we can also go backwards if someone needs!

PR #652, opened 2026-05-14
fix(dispatcher): honour `:causation_id` dispatch option on persisted events

### LINKED ISSUE #624: Causation ID should point from event to event, not commands
I use the Commanded library including directly reading events from the underlying EventStore. Due to the way Commanded implements `causation_id`, I find that I cannot track the chain of causation from event to event for multi-event chains. This is due to the fact that by default, commands are not persisted, and even if the command audit middleware is used, you would have to stitch together data from 2 separate data stores to establish a chain of causation.

In short, I would like to make a case for changing this behavior or allowing a configuration flag to do so.

**Conceptual / Philosophical Case**

Commands, whether kept in an audit log are not, are intended to be ephemeral entities in a CQRS/ES system. You should not have to replay commands (or retain commands at all once processed) to reconstruct a chain of causation, only events.

**Use-As-Intended Case**

`causation_id` is a concept in [EventStore](https://hexdocs.pm/eventstore/EventStore.html) that is bubbled up by Commanded. The EventStore docs say:
> causation_id - an optional UUID identifier used to identify which message you are responding to.

Note that the term "message" is used throughout the EventStore docs to refer specifically to a stored event. In other words, `causation_id` should be an event ID, not an ID of something external to EventStore. Setting it to something outside EventStore makes it difficult to use the powerful subscription and replay abilities that EventStore provides.

**Practical Case**

The current implementation makes it impossible to track chain of causation without persisting commands. Persistence of commands should be optional, and the decision to do so should be driven by the need to audit commands specifically, not required to make events audit-able. Even if commands are persisted, they are done so separately from EventStore making the implementation complexity of tracking chain of causation high.

Currently, I work around this by passing an event ID in the command dispatched in response to an event, then setting it as a field in any events dispatched as a result. ~~This means I cannot use ProcessManagers (but rather must use a "naked" event handler) because Process Managers do not allow access to event metadata.~~ Never mind, I see there is a version of handle that receives metadata.

**Suggestions**

* Allow a configuration flag to bypass command IDs when setting causation IDs. In this mode, events not caused by other events have no causation ID. Events emitted in response to commands dispatched from command processors set their causation ID to the ID of the event that caused the command dispatch.
* Provide access to command metadata in aggregates and command handlers, ~~as well as event metadata in process managers.~~
* Alternately provide the ability to configure what `causation_id` is set from. I could configure it to set `causation_id` for an event from a command's `causation_id` as opposed to from its id.

Thank you!

apre:
disclaimer: most of this PR is generated with claude code. I might not understand the deep implications of this changes, but the essential pouint is that it fixed my issue and seems to close the issue #624 at the same time.

Thank you for your time !

Closes #624.

`Commanded.Commands.Dispatcher.to_execution_context/2` previously hardcoded `ExecutionContext.causation_id` to the dispatched command's `command_uuid`, silently dropping any `:causation_id` supplied by the caller. As a result the `causation_id` persisted on events always pointed at a command — which is not stored in the event store by default — making it impossible to reconstruct a chain of causation from events alone for multi-event flows.

Read the caller-supplied `:causation_id` from the pipeline and fall back to `command_uuid` when it is not set. Default behaviour is unchanged when no `:causation_id` is supplied.

This also fixes a latent bug where process managers (`process_manager_instance.ex:229`) and event handlers already passed the handled event's id as `:causation_id`, but the value was thrown away before persistence. Events emitted by PM- or handler-dispatched commands now carry the handled event's id as `causation_id`, one hop per dispatch.

  - lib/commanded/commands/dispatcher.ex: read `pipeline.causation_id`, use `causation_id || command_uuid` on the ExecutionContext
  - lib/commanded/commands/router.ex: document the option on `dispatch/2`
  - test/commands/correlation_causation_test.exs: rename default-fallback test for clarity; assert explicit `:causation_id` lands on the event; flip the handler test (which was codifying the bug); add a PM-step assertion (one hop: MoneyTransferRequested → MoneyWithdrawn → MoneyDeposited)
  - CHANGELOG.md: entry under Unreleased / Bug fixes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,21 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
+
+* The `:causation_id` dispatch option is now honored on persisted events.
+  Previously it was silently overridden by the dispatched command's
+  `command_uuid`, which made chain-of-causation impossible to reconstruct from
+  events alone for multi-event flows (closes
+  [#624](https://github.com/commanded/commanded/issues/624)). When
+  `:causation_id` is not supplied, the command's own `command_uuid` is still
+  used as the fallback — so default behaviour is unchanged. This also fixes a
+  latent bug where process-manager- and event-handler-dispatched commands
+  passed the originating event's id as `:causation_id` (see
+  `process_manager_instance.ex:229` and `Commanded.Event.Handler` examples),
+  but it was being thrown away before persistence.
+
 ## v1.4.9
 
 ### Enhancements
--- a/lib/commanded/commands/dispatcher.ex
+++ b/lib/commanded/commands/dispatcher.ex
@@ -179,7 +179,7 @@ defmodule Commanded.Commands.Dispatcher do
 
     %ExecutionContext{
       command: command,
-      # honour caller-supplied `:causation_id`; fall back to command_uuid (#624)
+      # honour caller-supplied `:causation_id`; fall back to command_uuid
       causation_id: causation_id || command_uuid,
       correlation_id: correlation_id,
       metadata: metadata,
--- a/lib/commanded/commands/dispatcher.ex
+++ b/lib/commanded/commands/dispatcher.ex
@@ -160,7 +160,12 @@ defmodule Commanded.Commands.Dispatcher do
   end
 
   defp to_execution_context(%Pipeline{} = pipeline, %Payload{} = payload) do
-    %Pipeline{command: command, command_uuid: command_uuid, metadata: metadata} = pipeline
+    %Pipeline{
+      command: command,
+      command_uuid: command_uuid,
+      causation_id: causation_id,
+      metadata: metadata
+    } = pipeline
 
     %Payload{
       correlation_id: correlation_id,
@@ -174,7 +179,8 @@ defmodule Commanded.Commands.Dispatcher do
 
     %ExecutionContext{
       command: command,
-      causation_id: command_uuid,
+      # honour caller-supplied `:causation_id`; fall back to command_uuid (#624)
+      causation_id: causation_id || command_uuid,
       correlation_id: correlation_id,
       metadata: metadata,
       handler: handler_module,
--- a/lib/commanded/commands/router.ex
+++ b/lib/commanded/commands/router.ex
@@ -214,7 +214,7 @@ defmodule Commanded.Commands.Router do
       :ok = BankApp.dispatch(command, causation_id: some_event_id)
 
   When `:causation_id` is not provided, the command's own `command_uuid` is
-  used as the resulting events' `causation_id` — preserving prior behaviour.
+  used as the resulting events' `causation_id`.
 
   Process managers and event handlers that dispatch follow-up commands
   already propagate the handled event's id this way, so the full causation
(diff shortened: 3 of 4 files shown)


test/commands/correlation_causation_test.exs
```diff
@@ -88,6 +98,17 @@ defmodule Commanded.Commands.CorrelationCasuationTest do
       transfer_requested = EventStore.stream_forward(BankApp, transfer_uuid) |> Enum.at(0)
       [_, causation_id, _] = CommandAuditMiddleware.dispatched_commands(& &1.causation_id)
       assert causation_id == transfer_requested.event_id
+
+      # events emitted by a PM-dispatched command carry the handled event's
+      # id as their `causation_id` — one hop per PM step (#624)
```
TylerPachal: What do you mean by "one hop per PM step" here?

Also, no need to mention the PR number here:

```suggestion
      # id as their `causation_id` — one hop per PM step
```

lib/commanded/commands/router.ex
```diff
@@ -204,6 +204,22 @@ defmodule Commanded.Commands.Router do
...
+
+  You can associate a `:causation_id` with the dispatched command, which will
+  also be persisted as the `causation_id` of every event the command produces.
+  This lets you reconstruct a chain-of-causation across multiple events
+  without persisting commands separately.
+
+      :ok = BankApp.dispatch(command, causation_id: some_event_id)
+
+  When `:causation_id` is not provided, the command's own `command_uuid` is
+  used as the resulting events' `causation_id` — preserving prior behaviour.
```
TylerPachal: No need to mention "prior behavior" because that will lose its meaning/reference over time. 

```suggestion
  used as the resulting events' `causation_id`.
```

lib/commanded/commands/dispatcher.ex
```diff
@@ -174,7 +179,8 @@ defmodule Commanded.Commands.Dispatcher do
 
     %ExecutionContext{
       command: command,
-      causation_id: command_uuid,
+      # honour caller-supplied `:causation_id`; fall back to command_uuid (#624)
```
TylerPachal: No need to mention the PR in the comments:

```suggestion
      # honour caller-supplied `:causation_id`; fall back to command_uuid
```

CHANGELOG.md
```diff
@@ -1,5 +1,21 @@
 # Changelog
 
+## Unreleased
+
+### Bug fixes
```
TylerPachal, line 5: @drteeth Do we usually generate these automatically? Or do we add them manually like this?

TylerPachal:
@apre thanks for your PR!  Also thanks for the AI disclaimer. I have left some small comments/questions.  We will also wait to hear back from @drteeth because he has more context than I do.

CHANGELOG.md
apre: (was added by claude BTW).  I try to follow https://keepachangelog.com/en/1.0.0/  , this entries may be a bit verbose

test/commands/correlation_causation_test.exs
apre: I am sorry for the slop. I rephrased that part. If you with I can also remove that comment completly.

drteeth (approved):

PR #654, opened 2026-05-21
Allow causation to be nil  nil

apre:
Disclaimer: AI generated code.  The fix seems to works on my side and fits my personal needs. 

-----

Adds support for causation_id: nil as an explicit dispatch option to mark resulting events as chain roots.

Previously, Keyword.get(opts, :causation_id) returned nil whether the option was absent or explicitly nil, so the dispatcher fell back to command_uuid in both cases — there was no way to express "this event has no cause."

:causation_id is now resolved at the router boundary via Keyword.fetch/2, distinguishing three cases:

- opt absent → fall back to command_uuid (legacy behaviour, unchanged)
- causation_id: <uuid> → use the uuid (unchanged)
- causation_id: nil → persist NULL (new)

Side benefit: the dispatcher's to_execution_context/2 drops the || command_uuid fallback and reads pipeline.causation_id directly. Middleware that inspects pipeline.causation_id continues to see a uuid-or-nil, never a sentinel.

Why

WHERE causation_id IS NULL now identifies chain roots without a self-join against the events table — significantly cheaper on large event stores.

Backwards compatibility

- Callers that never pass :causation_id get identical behaviour.
- Callers that pass a uuid get identical behaviour.
- The only new behaviour is opt-in via causation_id: nil.

### DIFF
--- a/lib/commanded/commands/router.ex
+++ b/lib/commanded/commands/router.ex
@@ -547,13 +547,7 @@ defmodule Commanded.Commands.Router do
           command_uuid = Keyword.get_lazy(opts, :command_uuid, &UUID.uuid4/0)
           consistency = Keyword.fetch!(opts, :consistency)
           correlation_id = Keyword.get_lazy(opts, :correlation_id, &UUID.uuid4/0)
-
-          # Resolve `:causation_id` at the router boundary so the fallback
-          # decision happens once and downstream middleware/handlers see the
-          # final value. Three cases:
-          #   - opt absent              -> fall back to command_uuid (legacy)
-          #   - `causation_id: <uuid>`  -> use the uuid
-          #   - `causation_id: nil`     -> use nil (chain root)
+          # `Keyword.fetch/2` distinguishes absence from explicit `nil`
           causation_id =
             case Keyword.fetch(opts, :causation_id) do
               {:ok, value} -> value
--- a/lib/commanded/commands/router.ex
+++ b/lib/commanded/commands/router.ex
@@ -436,7 +436,12 @@ defmodule Commanded.Commands.Router do
       Options:
 
         - `causation_id` - an optional UUID used to identify the cause of the
-          command being dispatched.
+          command being dispatched. When the option is omitted, the command's
+          own `command_uuid` is used (preserving prior behaviour). Pass
+          `causation_id: nil` explicitly to mark the resulting events as chain
+          roots — they will be persisted with `causation_id` set to `NULL`,
+          letting `WHERE causation_id IS NULL` queries identify roots without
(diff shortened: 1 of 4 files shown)


lib/commanded/commands/router.ex
```diff
@@ -539,10 +544,16 @@ defmodule Commanded.Commands.Router do
           opts = Keyword.merge(@command_opts, opts)
 
           application = Keyword.fetch!(opts, :application)
-          causation_id = Keyword.get(opts, :causation_id)
           command_uuid = Keyword.get_lazy(opts, :command_uuid, &UUID.uuid4/0)
           consistency = Keyword.fetch!(opts, :consistency)
           correlation_id = Keyword.get_lazy(opts, :correlation_id, &UUID.uuid4/0)
+          # `Keyword.fetch/2` distinguishes absence from explicit `nil`
+          causation_id =
```
TylerPachal, line 551: No need for the AI's comments here :) 

```suggestion
          causation_id =
```

lib/commanded/commands/router.ex
```diff
@@ -436,7 +436,12 @@ defmodule Commanded.Commands.Router do
       Options:
 
         - `causation_id` - an optional UUID used to identify the cause of the
-          command being dispatched.
+          command being dispatched. When the option is omitted, the command's
+          own `command_uuid` is used (preserving prior behaviour). Pass
+          `causation_id: nil` explicitly to mark the resulting events as chain
+          roots — they will be persisted with `causation_id` set to `NULL`,
+          letting `WHERE causation_id IS NULL` queries identify roots without
+          a self-join.
```
TylerPachal, line 444: The term "chain root" is not something we have in the codebase today.  To keep this PR small I suggest focusing this comment on the behavior of the function call, and not the downstream motivations for using it (I assume your AI agent added this because you described the problem you wanted to solve and it documented it all here).

```suggestion
command being dispatched. When the option is omitted, the command's
own `command_uuid` is used. Pass `causation_id: nil` to explicitly 
set resulting event's `causation_id` to `NULL` in the the event 
store.
```

drteeth: I think it's reasonable to be suspicious of nil/null as a sentinel value. We recently added support for the Postgres eventstore to allow arbitrary strings as causation and correlation IDs. Would this cover that same need?

If you wanted to stay with UUIDs, maybe we could find a way of using UUIDv5 to have a known chain root UUID.

apre:
I close this PR following @drteeth  comment

> If you wanted to stay with UUIDs, maybe we could find a way of using UUIDv5 to have a known chain root UUID.

Thank you for the idea !

PR #665, opened 2026-07-26
Remove unused require Logger

wimwian:
## Summary

Removes the `require Logger` line in `Commanded.Commands.Dispatcher` (`lib/commanded/commands/dispatcher.ex`). The module never calls `Logger.*`, so on Elixir 1.20 this triggers an "unused require Logger" compiler warning. This change has no runtime effect — it just clears the warning.

I read through the entire file (including private functions and the nested `Payload` module) to confirm there is no `Logger` usage anywhere before removing the require.

## Test plan

- [x] `mix deps.get && mix compile --force` — warning gone, no new warnings introduced
- [x] `mix test` — 500 passed, 1 skipped, 1 excluded (pre-existing, unrelated to this change)
- [x] `mix format` — no changes needed beyond the diff


drteeth:
Ty!

